PR #85, opened 2026-07-25
hooks: record pr-review.sh executable in git

alxjrvs:
`boom`'s `mode = "755"` chmods the link **source**, so a hook committed 644 is re-chmodded on every `boom source` — leaving boom's config-repo clone permanently dirty.

That clone is what `~/.claude/settings.json` symlinks into, so the drift check added in #83 went red on its first reconcile and would have stayed red forever, training you to ignore it.

Caught by that check on its very first run, which is the argument for having it.


alxjrvs (commented):
**Scope.** `git diff @{upstream}...HEAD` is empty (branch is pushed and identical to `origin/fix/hook-exec-bits`). The real PR scope is `git diff origin/main...HEAD` = commit `561b540`, which is a single file-mode change (`dot-claude/hooks/pr-review.sh` 100644 → 100755), plus the uncommitted working-tree change, which is in scope per the instructions.

**Assessment of the committed change:** correct and complete. `pr-review.sh` was the sole holdout — `rebase-guard.sh`, `worktree-checkout-guard.sh`, `tests/run.sh`, `git-template/hooks/pre-commit`, and `hooks/op-agent.sh` were all already `100755` in the index, and the boomfile declares `mode = "755"` for this link (`boomfile.toml:180-182`), so boom was chmod'ing the source itself and leaving `~/.local/state/boom/config-repo` permanently dirty. Recording 755 in git is the right fix and resolves that. I checked for other scripts exec'd by path with a missing bit; there are none.

Findings:

- `dot-claude/hooks/pr-review.sh:1` — An uncommitted working-tree change flips the file back from mode 100755 to 100644, exactly reverting the PR's only change. `./dot-claude/hooks/pr-review.sh` exits 126 "permission denied" from this checkout (verified), and `dot-claude/settings.json:78,88` invokes the bare path, so the `PostToolUse` review silently never fires — the exact bug the commit fixes. A `git commit -a`/`git add -A` before merge would make the PR a no-op. Fix: `chmod 755 dot-claude/hooks/pr-review.sh` (or `git checkout --` it).

Note on non-findings I deliberately left out: `pr-review.sh`'s body is unchanged by this diff and was already live on the machine (boom's chmod made it executable at the link target), so latent nits in it — e.g. a per-SHA lock dir that is never reclaimed if the outer script dies between `mkdir "$lock"` and the background subshell, permanently suppressing re-review for that SHA — are pre-existing and out of this diff's scope.

PR #86, opened 2026-07-25
mcp: uninstall the GitHub server; delete op-agent's orphaned `header` verb

alxjrvs:
## The server

The GitHub MCP had **0 calls across all 3,410 session transcripts** and had been silently
failing to connect:

```
github: api.githubcopilot.com/mcp/ (HTTP) - ✘ Failed to connect
  — Incompatible auth server: does not support dynamic client registration
```

It carried a classic `repo`+`workflow` SSO-authorized PAT into every session in every repo
for no measured value — the largest privilege-per-value item in the setup. Removed from the
user scope **and** the `/Users/jarvis/Code/SU-SRD` project scope in `~/.claude.json`.

Worth noting for the record: the unquoted-`op://`-ref bug fixed in #83 was *a* bug here (the
helper had been emitting `{}`), but it was not *the* blocker — the header resolves correctly
now and the server still refused. The real cause was never found, and is now moot.

## The cascade

`op-agent header` existed solely to format this server's `headersHelper` Bearer line. With
the server gone it has **no consumer at all**. `op-agent.sh`'s own stated rule is that every
verb has a live consumer, so it goes rather than sitting "in case" — a dead verb that
resolves a PAT is strictly worse than no verb. Restore from git history if an HTTP MCP
server that bearer-authenticates is ever added back.

`op-agent` is now `secret` / `git-credential` / `provision` / `status`, all four with live
consumers.

## Docs

Both CLAUDE.mds asserted the `headersHelper` wiring as current fact in four places. Corrected.
The "every verb has a live consumer" line is now framed as **the rule it is, enforced by
deletion** — it had quietly gone false while reading as a description.

Also recorded: zero calls means *broken or unused*, and those are indistinguishable from usage
data alone. Two of three configured servers turned out to be broken, not unused — check
`claude mcp list` before concluding anything is dead weight.

## Verified

- `op-agent status` still OK; `secret` / `git-credential` / `provision` intact; agent pushes work
- `header` now falls through to usage
- shellcheck + shfmt clean; 33/33 guard tests pass
- **`boom verify` is green** — the MCP health check from #83 had been failing on this server


alxjrvs (commented):
Reviewed the single-commit diff (`0fd4aaf`, 3 files: `CLAUDE.md`, `dot-claude/CLAUDE.md`, `hooks/op-agent.sh`) plus a sweep of the rest of the repo and the live `~/.claude.json` for consumers of the deleted verb.

**The headline: the PR's core premise is wrong.** `header` was not orphaned — a `render` HTTP MCP server is still configured with a `headersHelper` that execs it.

Findings:

- `hooks/op-agent.sh:18` — **Deleted `header` verb still has a live consumer.** `~/.claude.json` has, under project `/Users/jarvis/Code/SU-SRD` (dir exists): `"render": {"type":"http","url":"https://mcp.render.com/mcp","headersHelper":"/Users/jarvis/.local/bin/op-agent header \"op://claude-agent/Render API Key/credential\""}`. `boomfile.toml:444-446` symlinks `hooks/op-agent.sh` to exactly that path, so after merge + `boom source` the call falls through to `*)` at line 170 → usage on stderr, `exit 2`, no JSON (the old `cmd_header` always emitted a well-formed `{}`). Render MCP fails to connect in that repo, and `boom verify`'s own `claude mcp list | grep '✘'` check (boomfile.toml:260) starts failing there. Tellingly, the CLAUDE.md line this PR deleted read verbatim "HTTP MCP servers (**GitHub, Render**) → `headersHelper` formatted from `op-agent header`".
- `CLAUDE.md:91` — "No HTTP MCP server is configured today — the GitHub MCP, the only one that used a `headersHelper`" is factually false, and line 106 tells a future reader to restore `header` only "if an HTTP MCP server is ever added" — so the render breakage stays invisible.
- `.zprofile:13` (and `zsh/00-exports.zsh:14`) — stale `op-agent header` / GitHub-MCP references the PR missed; the repo's docs now contradict each other.
- `dot-claude/CLAUDE.md:14` — editing artifact: "...confused-deputy**, ;** the user-scope GitHub MCP...".

Suggested fix: keep `cmd_header` (it has a consumer), and narrow the docs to "the GitHub MCP was removed; render still uses `headersHelper`" — or, if render is genuinely meant to go too, remove that entry from `~/.claude.json` in the same change and verify with `claude mcp list` from `~/Code/SU-SRD`.

PR #93, opened 2026-07-28
statusline: migrate the upstream to thegnarco/claude-statusline

alxjrvs:
Migrates the Claude Code statusline from `alxjrvs/claude-statusline` to the collectively-owned [`thegnarco/claude-statusline`](https://github.com/thegnarco/claude-statusline).

## This is a provenance change, not a behavior change

The Gnar repo was **seeded from `alxjrvs/claude-statusline@7107dc5`** — the exact commit installed on this machine. At the switch:

| | |
|---|---|
| `statusline.sh` | **byte-identical** (`md5 6be767192c15195cf724c3ae0f555939`, 866 lines) |
| `subagent-statusline.sh` | **byte-identical** (`md5 3a60f8244292f04b04e4b85cd623ec33`) |
| `install.sh` | **byte-identical** — same `~/.local/bin/claude-statusline` + `claude-subagent-statusline` symlink names |

So nothing renders differently today, `dot-claude/settings.json` needs **no edit at all**, and what actually changes is where future updates come from.

Worth noting the road not taken: the gnar statusline is *also* vendored into the `toolkit@gnar` plugin (`/gnar-statusline`), but that path would mean enabling a whole plugin for one script and copying an untracked `~/.claude/gnar-statusline.sh` outside boom's declarative link contract — and the plugin's vendored copy is already stale (19k, vs the repo's 36k). The standalone repo keeps the existing hook shape, so the boomfile stays the single source of truth.

## The hook needed a real fix

Changing one line of TOML would have converged **nothing**. The hook only honored `repo=` on a *fresh* clone; thereafter it `pull`ed whatever remote the checkout already had — so every existing machine would have kept tracking `alxjrvs` forever, silently, while looking healthy.

It now re-clones when the declared upstream differs. **Re-clone rather than `remote set-url` + pull**, because the repo was seeded rather than forked: `7107dc5` does not exist in it (`HTTP 422` from the compare API), the histories are unrelated, and no `pull --ff-only` can cross that boundary. The checkout is derived state that is never edited in place, so discarding it is free.

`verify` now also reports upstream drift — without it, a stale machine keeps pulling the old remote while the scripts on PATH still look fine, exactly the quiet divergence `boom verify` exists to name.

## Verification

Ran the hook end-to-end against a **sandboxed `HOME`**, from a pre-seeded old-upstream checkout — the live setup was confirmed untouched throughout:

- `verify` before → warns on both the missing bin and the wrong upstream
- `sync` → `upstream is now github.com/thegnarco/claude-statusline — re-cloning` → clones → runs `install.sh`
- resulting remote → `https://github.com/thegnarco/claude-statusline.git`; `verify` after → clean
- **second run pulls instead of re-cloning** (idempotent)
- the resulting script renders correctly against sample statusline JSON

`biome check`, `taplo check`, and `gitleaks` all pass (full lefthook pre-commit).

## Follow-up on merge

This machine still points at the old remote until boom re-sources from the merged `main`:

```
boom source --only=claude_statusline
```

Until then `boom verify` will warn about the drift — that's the new check doing its job, not a failure.


alxjrvs (commented):
Reviewed the migration of the statusline upstream to `thegnarco/claude-statusline`. The boomfile/README/CLAUDE.md prose changes are accurate (I verified `TheGnarCo/claude-statusline` exists, is public, clones fine at the lowercase URL the boomfile pins, and ships a byte-identical `install.sh`). All findings are in the hook's new re-clone logic.

Two things I checked and deliberately did **not** report: `api.with` is populated for every verb (boom's `reconcileHook` builds `with: entry.with ?? {}` regardless of verb), so `api.with.repo` in `verify()` can't throw; and the `rm -rf` target resolves to boom's derived clone at `~/.local/state/boom/claude-statusline`, not your dev checkout at `~/Code/DevEnv/claude-statusline` (which does hold two unpushed branches and a locked worktree) — so the "derived state, never edited in place" claim in the comment holds.

Findings:

- `hooks/claude_statusline.ts:46` — `rm -rf` runs before a `.nothrow().quiet()` clone; the on-PATH `~/.local/bin/claude-*statusline` symlinks point into that directory, so a failed clone (offline/5xx) leaves them dangling and breaks the statusline in every session, while sync still prints `ok("statusline cloned")`. Previously the same failure was a harmless no-op pull. Clone to a temp dir and swap.
- `hooks/claude_statusline.ts:46` — the `rm`'s exit status is discarded; if the delete fails, `existsSync(.git)` is still true so the hook pulls and reinstalls the OLD upstream and reports "statusline updated" — green exit, migration silently not performed.
- `hooks/claude_statusline.ts:36` — the `dryRun` early return precedes the new re-clone block, so `boom source --dry-run` never discloses that an existing checkout will be deleted.
- `hooks/claude_statusline.ts:27` — `pointsElsewhere` substring-matches the whole `.git/config` instead of `remote.origin.url`: a second remote or `pushurl` mentioning the new URL makes it return false (keeps pulling the stale origin forever, verify reports clean), while an SSH remote, a no-`.git` clone, or the repo's real capitalization (`TheGnarCo` vs the pinned `thegnarco`) makes it return true and destructively re-clone an already-correct checkout.

PR #95, opened 2026-07-29
statusline: use the canonical TheGnarCo casing for the upstream repo

alxjrvs:
## What

Spell the Gnar statusline upstream the way GitHub spells it: `TheGnarCo/claude-statusline`, not `thegnarco/claude-statusline`. Six references, one string each — `boomfile.toml`'s `[section.hook.with] repo`, the `DEFAULT_REPO` constant in `hooks/claude_statusline.ts`, and four doc mentions (`README.md`, `CLAUDE.md`, `dot-claude/CLAUDE.md`, `dot-claude/DECISIONS.md`).

`gh api repos/thegnarco/claude-statusline` returns `full_name: TheGnarCo/claude-statusline`.

## Why it's more than cosmetic

GitHub resolves either casing, so nothing was broken. But `claude_statusline.ts`'s `pointsElsewhere()` decides whether to re-clone by substring-matching the configured URL against the remote recorded in the checkout's `.git/config` — and git stores whatever URL it was cloned with. Feeding that comparison the canonical spelling keeps it resting on the real name rather than on a redirect.

Note this costs one extra re-clone on any machine that already cloned via the lowercase URL: its `.git/config` won't match the new string, so the next `boom source` re-clones and thereafter matches. That checkout is derived state (never edited in place), so the re-clone is free of consequence.

## Separate, not fixed here: the migration hasn't reached this machine, and `--only` won't land it

The live checkout (`~/.local/state/boom/claude-statusline`, what `~/.local/bin/claude-*statusline` symlinks into) still has `origin = alxjrvs/claude-statusline`. An earlier revision of this description said `boom source --only=claude_statusline` would land it. **That is wrong**, and the reason is worth recording:

`~/.local/state/boom/config-repo` — boom's own clone of this repo, the one it actually reads `boomfile.toml` and `hooks/` from — is parked on the feature branch `claude/enable-binfinite-context-globally`, 3 commits behind `origin/main` and 1 ahead. Those 3 missing commits include #93, the statusline migration. So its `boomfile.toml` still declares `repo = "github.com/alxjrvs/claude-statusline"` and its copy of the hook has no `pointsElsewhere`/re-clone logic at all. `boom source` therefore does exactly what it is told: pulls `alxjrvs` and reports `statusline updated`. The new code is fine; boom is running old code.

The one commit ahead is a pre-squash version of what landed as #92 — `git patch-id --stable` is identical for both (`f843779b…`), so moving that clone back to `main` loses nothing. Getting `config-repo` current is the actual fix; the hook's own `rm -rf` + re-clone then swaps the upstream on the next sync.

(A second, unrelated checkout at `~/Code/DevEnv/claude-statusline` is an orphan — nothing symlinks to it and it is not the hook's `TARGET`, since boom resolves `import.meta.dir` inside `config-repo/hooks`.)

## Verification

- `taplo check --no-schema $(git ls-files '*.toml')` — pass
- `biome check` — pass
- lefthook pre-commit (toml-validity, biome-check, gitleaks) — pass

(`taplo fmt --check` reports `boomfile.toml` unformatted, but that is pre-existing on `main` and is not a CI gate — CI runs `taplo check`.)

https://claude.ai/code/session_01XrhtRSdazyoVnJZcEqy35T


alxjrvs (commented):
Reviewed `main...HEAD` (commit 4c0fda1) — a casing-only change of the statusline upstream from `thegnarco/claude-statusline` to `TheGnarCo/claude-statusline` across `boomfile.toml`, `hooks/claude_statusline.ts`, and four docs. The casing is canonically correct (`gh api repos/thegnarco/claude-statusline` → `full_name: TheGnarCo/claude-statusline`), but the hook's drift detection is case-sensitive, so the cosmetic change has a real behavioral effect.

Findings:

- `/Users/jarvis/Code/DevEnv/dotFiles/.claude/worktrees/statusline-upstream-casing/hooks/claude_statusline.ts:27` — `pointsElsewhere` uses a case-sensitive `readFileSync(...).includes(url)`, so a machine already cloned from the lowercase URL is judged to point at a *different* upstream: `sync()` `rm -rf`s a current checkout of the same repo and re-clones, and `boom verify` falsely warns "statusline upstream is not TheGnarCo/claude-statusline" until then. The re-clone's own justification ("a re-seeded upstream shares no history") doesn't apply to a casing edit. Lowercasing both sides of the compare fixes it.
- `/Users/jarvis/Code/DevEnv/dotFiles/.claude/worktrees/statusline-upstream-casing/hooks/claude_statusline.ts:46` — the newly-triggered path does `rm -rf ${TARGET}` then `git clone` with `.nothrow().quiet()` and no exit-code check; if the clone fails the checkout is destroyed while the hook still emits `ok("statusline cloned")` and `verify()` still says "statusline on PATH" (the previously installed binary survives), so the failure is silent.

Notes (not filed as bugs): this machine's checkout is still on `alxjrvs/claude-statusline`, so it will re-clone regardless — the finding above bites machines that synced after #93. `dot-claude/CLAUDE.md:68` and `dot-claude/settings.json:116` still carry lowercase `thegnarco/agent-skills`; that's a different repo and settings.json is client-written, so leaving it is defensible, but the doc/global-CLAUDE.md casing now disagrees.

PR #96, opened 2026-07-31
feat(gh): adopt stacked PRs via the official github/gh-stack extension

alxjrvs:
Makes stacked PRs the standing preference for agent work, and installs the tool that makes them executable.

## Why now

GitHub put stacked PRs into [public preview on 2026-07-30](https://github.blog/changelog/2026-07-30-stacked-pull-requests-are-now-in-public-preview/) with a first-party CLI extension. That is what changed the answer. Stacking was always the right shape for agent output — many small reviewable layers instead of one 40-file PR — but every prior implementation was a third-party tool (Graphite, `git-branchless`, four community `gh-stack` forks) with its own metadata model, hosted service, or both, and **native over special** ruled them all out. A GitHub-shipped extension that stores state in `.git/gh-stack` and drives the stock PR API *is* the stock behavior.

## What changed

- **`boomfile.toml`** — a new `gh extensions` section installs `github/gh-stack` install-if-absent.
  - It **must** sit after `packages`: `gh` comes from mise and sections run in file order, so a fresh machine has no `gh` on PATH before that.
  - The grep matches the `github/` owner deliberately — `gh ext search stack` returns four same-named community extensions and only one is GitHub's.
  - Imperative by necessity: `gh` has no declarative manifest and boom has no `gh` package manager (brew/mise/cargo/npm/pipx/gem/apt/dnf/flatpak only). Same shape as the existing Claude CLI install step. No `|| true` — a missing extension should fail loudly.
- **`agent-friendly-repo` skill** — stacked PRs added to the checklist and the completion path.
- **`dot-claude/CLAUDE.md`** — the workflow rule, folded into the standing merge preference.
- **`dot-claude/DECISIONS.md`** — the reasoning, per the contract/decision-record split.
- **`CLAUDE.md`** — gh extensions noted as the exception to the brew-vs-mise packaging policy.

## Two things that fell out of reading the tool's actual contract

- **The existing checklist was already a *precondition*, not merely compatible.** `gh stack modify` refuses a history with merge commits or diverged branches, and `gh stack merge` states outright that bypassing merge requirements is **not supported** for stack merges. So `required_linear_history` and `bypass_actors: []` — already in the skill on other grounds — are exactly what stacks need. A repo whose agent path leans on a bypass actor cannot use stacks at all.
- **The rebase-guard does not see `gh stack`.** It tokenizes for `git push` / `gh pr create` specifically, so `gh stack submit`/`push`/`sync` sail past it. Deliberately *not* fixed by widening the guard: `gh stack sync` already does fetch + cascading rebase + push natively, and re-deriving stack-aware behind-ness across N branches would duplicate that in security-relevant shell. The rule lives in prose, with the caveat that prose is advisory stated plainly.

Also flagged and **not** adopted here: `gh skill install github/gh-stack --agent claude-code` writes a GitHub-authored skill into `~/.claude/skills/` — the same directory boom glob-links into. That is a new agent-context surface from outside the repo, and the doctrine is that those get enumerated before adoption, not installed as a side effect.

## Verification

- `taplo check --no-schema` on all 7 tracked TOMLs — clean; section ordering asserted (`packages` idx 5 < `gh extensions` idx 6).
- `gitleaks detect` — no leaks.
- `dot-claude/hooks/tests/run.sh` — 33/33 guard regression tests pass.
- The `run` step was executed twice on this machine: installs `gh stack v0.1.0`, then no-ops (exit 0 both times).

https://claude.ai/code/session_01LE5VUNcrTgBypsmYcD6doG

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -62,6 +62,8 @@ Small steps (`chmod 700 ~/.ssh`, `lefthook install`) are inline `run` (`on = "ap
 
 `Brewfile` holds **only** `mise` (bootstrap), casks (GUI apps, fonts), and system libs with no mise equivalent. `mise.toml` holds all language toolchains AND dev CLIs (`jq`, `shellcheck`, `shfmt`). If you're about to add a CLI to `Brewfile`, stop — it goes in `mise.toml` unless it's `mise` itself or a cask.
 
+**`gh` extensions are the one exception to both** — they aren't packages either manager knows about, and boom has no `gh` manager, so the `gh extensions` section carries install-if-absent `run` steps (today: the official `github/gh-stack`, for stacked PRs). That section **must stay after `packages`**: `gh` comes from mise and sections run in file order, so a fresh machine has no `gh` on PATH before it.
+
 ## Terminal: Ghostty (sole terminal)
 
 Ghostty is the only terminal (`TERMINAL=ghostty`, `ghostty/config`) — one symlinked file for rendering + keybinds/visor. Parallel Claude Code sessions run through `claude agents` mode directly (standalone Ghostty, no multiplexer: `claude agents` already does parallel dispatch, session persistence, and git-worktree isolation, so a multiplexing terminal is redundant). `boom code` mirrors `~/Code` into workspaces: `boom code init [DIR]` records the dir, `boom code claude` symlinks every repo into one flat dir and opens `claude agents` there so each repo is `@`-taggable for dispatch. (cmux, a libghostty agent multiplexer, was removed for this reason; `boom code cmux` — the per-repo workspace variant — is an unused engine feature.)
--- a/boomfile.toml
+++ b/boomfile.toml
@@ -296,9 +296,17 @@ name = "gh extensions"
 # file order, so a fresh machine has no `gh` on PATH until packages has run.
 #
 # `gh` has no declarative manifest and boom has no `gh` pkg manager (brew/mise/cargo/npm/
-# pipx/gem/apt/dnf/flatpak only), so this is imperative by necessity: install-if-absent, then
-# let gh's own daily extension auto-update keep it current. Not `|| true` — an extension the
-# stacked-PR workflow depends on should fail loudly, not silently go missing.
+# pipx/gem/apt/dnf/flatpak only), so this is imperative by necessity: install-if-absent. Not
+# `|| true` — an extension the stacked-PR workflow depends on should fail loudly, not silently
+# go missing.
+#
+# Deliberately install-only, NOT upgrade. gh does not auto-update extensions: it "will check
+# for new versions once every 24 hours and display an upgrade notice" when the extension runs
+# — a notice, not an install. So this pins nothing and refreshes nothing; `gh stack` stays on
+# whatever version first landed until someone runs `gh extension upgrade gh-stack`. That is a
+# live trap for a v0.1.0 preview tool, and the daily notice is the only thing that surfaces it.
+# Revisit (add an upgrade step, or `boom lock`-style pinning) if the version drifts far enough
+# to bite; don't assume this section keeps it current, because it doesn't.
 #
 # github/gh-stack is the OFFICIAL GitHub extension (public preview since 2026-07-30), which
 # matters because `gh ext search stack` returns four same-named community forks
--- a/boomfile.toml
+++ b/boomfile.toml
@@ -290,6 +290,30 @@ pkg = [
 ]
 
 [[section]]
+name = "gh extensions"
+
+# MUST stay after the `packages` section — `gh` itself comes from mise, and sections run in
+# file order, so a fresh machine has no `gh` on PATH until packages has run.
+#
+# `gh` has no declarative manifest and boom has no `gh` pkg manager (brew/mise/cargo/npm/
+# pipx/gem/apt/dnf/flatpak only), so this is imperative by necessity: install-if-absent, then
+# let gh's own daily extension auto-update keep it current. Not `|| true` — an extension the
+# stacked-PR workflow depends on should fail loudly, not silently go missing.
+#
+# github/gh-stack is the OFFICIAL GitHub extension (public preview since 2026-07-30), which
+# matters because `gh ext search stack` returns four same-named community forks
+# (boneskull/, VladimirAnaniev/, ThePlenkov/gh-stackx, 134130/gh-domino). The `github/` owner
+# in the grep is load-bearing: it is what distinguishes the real one from a fork.
+#
+# Why it earns a place: stacked PRs are the standing preference for agent work (see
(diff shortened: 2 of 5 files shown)


alxjrvs (commented):
Reviewed `main...HEAD` (2 commits: 80a3b8d + 4c897f6, which landed mid-review and already fixed the "gh auto-updates extensions" claim I'd flagged). The diff is one executable change (a boomfile `run` step) plus prose; I verified the prose against the actual installed `github/gh-stack` v0.1.0 binary (`--help` output and embedded strings), the `gh` binary, and the repo's own hooks/settings.

**Findings (8):**

- `dot-claude/CLAUDE.md:156` — The documented recipe (`init → add → submit → sync`) omits `--open`. `gh stack submit` in a non-interactive shell (i.e. every agent Bash call) creates **draft** PRs; `gh stack merge` refuses drafts (`pull request #%d is a draft; mark it ready for review before merging`) and auto-merge can't be enabled on them. The agent reports success and nothing lands.
- `dot-claude/skills/agent-friendly-repo/SKILL.md:44` — "auto-merge coexists with stacks" is false. The binary contains `DisableAutoMerge` and prints `Disabled auto-merge for PR %s (incompatible with stacked PRs)`. Stacking a PR turns off the repo's standing completion path (`gh pr merge --auto --squash`); the `unstack` behavior cited as evidence is the *consequence*, not a counterexample.
- `dot-claude/CLAUDE.md:159` — The diff carefully analyses the rebase-guard's blindness to `gh stack` but misses the identical gap in the pr-review `PostToolUse` hook: settings.json gates on `Bash(gh pr create:*)`/`Bash(git push:*)` and pr-review.sh re-gates on the same two strings. `gh stack submit` matches neither, so stacked PRs silently lose the adversarial review + `claude-review` status.
- `boomfile.toml:322` — Fresh-machine bootstrap breaks: mise installs `gh` but nothing authenticates it (verified: unauthenticated gh refuses all API calls), so `gh extension install` fails, and with no `|| true` the sync halts before ~10 later sections (macOS defaults, SSH perms, lefthook, op-agent provision, launchd, git-signing…).
- `boomfile.toml:322` — A pre-existing same-named community `gh-stack` makes the owner-qualified grep miss *and* the install abort permanently (`there is already an installed extension that provides the "stack" command`) — every `boom source` fails forever, with no self-heal.
- `dot-claude/skills/agent-friendly-repo/SKILL.md:39` — Ruleset `required_linear_history` (what may merge into main) is conflated with `gh stack modify`'s check on *local branch* history (`modify requires linear history`). Neither implies the other; step 9's "stack readiness" report measures the wrong thing.
- `dot-claude/CLAUDE.md:164` — "Stack merges cannot bypass merge requirements" ⇒ bypass actors are *irrelevant*, not forbidden. Calling `bypass_actors: []` a stack precondition (also SKILL.md:42, :156) could drive an unnecessary ruleset mutation.
- `boomfile.toml:312` — "four same-named community forks" is wrong on both count and kind: only `boneskull/gh-stack` and `VladimirAnaniev/gh-stack` share the name; `gh-stackx`/`gh-domino`/`gh-chain` differ; none are forks. Repeated in SKILL.md:34 and DECISIONS.md.

Things I checked that hold up: section ordering after `packages` is correct in-file; the grep matches real `gh extension list` output; `gh stack` subcommand names, `.git`-scoped (per-worktree, so genuinely not inherited) stack state, all-or-nothing merge semantics, and the merge-queue caveat are all accurate; the rebase-guard really does tokenize only `git push`/`gh pr create`, so that claim is right.

alxjrvs (commented):
Reviewed the `gh-stack` adoption PR (5 files: `boomfile.toml` run step + docs in `CLAUDE.md`, `dot-claude/CLAUDE.md`, `DECISIONS.md`, `agent-friendly-repo/SKILL.md`). Since this is a docs-plus-one-shell-command change, I verified the factual claims directly against the installed `github/gh-stack` v0.1.0 (its `--help` text and binary strings), against `gh` 2.94.0's behavior, and against `dot-claude/hooks/rebase-guard.sh`. The rebase-guard claim, the `.git/gh-stack` per-clone claim, and the all-or-nothing/merge-queue claims all check out; the eight below do not.

8 findings:

- `boomfile.toml:322` — `gh extension install` requires an authenticated `gh` (confirmed: errors with "please run: gh auth login" under an empty config), and boom never provisions `gh auth`. With no `|| true`, the documented fresh-machine bootstrap (`curl … | sh && boom source set alxjrvs/dotFiles`) fails at this section. The "must stay after `packages`" comment only addresses PATH — and even that is shaky, since mise's shim dir enters PATH via the `.zshenv` boom just linked, not via the running boom process.
- `dot-claude/CLAUDE.md:156` — the prescribed recipe stops at `gh stack submit`, which in a non-interactive terminal (i.e. every agent) creates **draft** PRs with auto-generated titles unless `--open` is passed. Drafts are rejected by `gh stack merge` ("open and not a draft") and by `gh pr merge --auto`, so the documented agent path produces an unmergeable stack. `--open` is never mentioned.
- `dot-claude/CLAUDE.md:163` — "cannot bypass merge requirements → so it needs empty `bypass_actors`" is a non-sequitur; the extension never bypasses, which says nothing about the repo's ruleset. Repeated as a hard precondition at `SKILL.md:41` and `DECISIONS.md:274`, and it drives the step-9 readiness report at `SKILL.md:137`.
- `dot-claude/skills/agent-friendly-repo/SKILL.md:39` — `required_linear_history` (a rule about what lands on the default branch) is conflated with `gh stack modify`'s check on the stack's own local history ("modify requires linear history" / "<branch> contains merge commits", both confirmed in the binary). Wrong in both directions as a readiness gate.
- `boomfile.toml:322` — a pre-existing same-named community extension (`boneskull/gh-stack`, `VladimirAnaniev/gh-stack`) defeats the owner-qualified grep and then fails the install with "there is already an installed extension that provides the \"stack\" command", wedging every subsequent `boom source` with no remediation hint.
- `dot-claude/CLAUDE.md:163` — `gh stack merge` run non-interactively merges the whole stack immediately using the last-used merge method, which is not the CI-gated `gh pr merge --auto` path the surrounding doctrine mandates, and is not covered by the enumerated `Bash(gh pr merge:*)` allow entry.
- `dot-claude/skills/agent-friendly-repo/SKILL.md:60` — "`gh stack sync` prunes merged branches locally (`--prune` to skip the prompt)" inverts the flag: sync never prunes by default and there is no prompt; `--prune` enables deletion.
- `boomfile.toml:312` — "four same-named community forks" is inaccurate; only two of the four named share the `gh-stack` name (`gh-stackx` and `gh-domino` are different commands), and `harsh183/gh-chain` is omitted. Same figure repeated at `DECISIONS.md:251`. The owner-in-grep rationale still holds, just on two rather than four.

PR #97, opened 2026-08-03
feat(ci): auto-merge Dependabot's minor/patch action bumps

alxjrvs:
## What

Two new files plus the doc trail for them:

- `.github/dependabot.yml` — `github-actions` ecosystem only, weekly, minor+patch grouped into one PR. Majors are excluded from the group, so they arrive individually and stay manual.
- `.github/workflows/dependabot-auto-merge.yml` — one job, gated on `github.actor == 'dependabot[bot]'`, that calls `gh pr merge --auto --squash`. It merges nothing itself; `lint` + branch protection stay the gate.
- `agent-friendly-repo` skill — a new "Dependabot auto-merge" checklist block, procedure step 9, and a guardrail, so this is reusable rather than one-off.
- `dot-claude/CLAUDE.md` + `DECISIONS.md` — the rule and the why.

## Why a workflow at all

There is no native switch. `dependabot.yml` has no automerge key (Renovate does; Dependabot doesn't), and the per-PR auto-merge button needs a human click. What the agent-friendly checklist *already* provides is the precondition: **no required human review**, so a Dependabot PR needs nothing but a green aggregate check.

## The four constraints

| Constraint | Consequence if ignored |
|---|---|
| `on: pull_request`, not `pull_request_target` | Dependabot runs get a read-only token by default but have respected `permissions:` since Oct 2021 — `pull_request_target` would hand a write token to a base-context run for nothing |
| Actions secrets unavailable to Dependabot runs | A non-hermetic CI job fails every Dependabot PR and auto-merge silently never fires. `lint.yml` here is hermetic |
| `GITHUB_TOKEN` can't add a PR to a merge queue | This and the skill's optional merge-queue step are mutually exclusive without a PAT/App token |
| Allowlist (`== minor \|\| == patch`), never `!= major` | An empty `update-type` must fail closed to a manual PR, not auto-merge |

Majors stay manual on purpose: an action major changes what code runs against a write-scoped token, which is the supply-chain surface the *Standing threats* section exists to keep small.

## Verification

- Both YAML files parse; workflow shape confirmed (job `auto-merge`, correct `if`, `permissions: {contents: write, pull-requests: write}`).
- `biome check` clean, guard regression suite 33/33, gitleaks clean.
- The auto-merge path itself is unverified until Dependabot opens its first PR — nothing here can be exercised before then.

https://claude.ai/code/session_018TCXxDahKWEKqsAnuTCKTr

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,22 @@
+# Dependabot — github-actions only.
+#
+# This repo has no npm/bun manifest to update (boom is a binary, mise owns the
+# toolchains, brew owns the casks), so the pinned action tags in
+# .github/workflows/ are the entire dependency surface.
+#
+# minor+patch land grouped in one weekly PR, which .github/workflows/
+# dependabot-auto-merge.yml auto-merges once `lint` is green. Majors are
+# excluded from the group, so they arrive as individual PRs and stay manual —
+# an action major can change what code runs against a write-scoped token.
+version: 2
+updates:
+  - package-ecosystem: github-actions
+    directory: "/"
+    schedule:
+      interval: weekly
+    commit-message:
+      prefix: chore(actions)
+    groups:
+      actions:
+        patterns: ["*"]
+        update-types: ["minor", "patch"]
--- a/.github/workflows/dependabot-auto-merge.yml
+++ b/.github/workflows/dependabot-auto-merge.yml
@@ -1,0 +1,42 @@
+name: dependabot-auto-merge
+
+# Turns on GitHub's own auto-merge for Dependabot's grouped minor/patch action
+# bumps. It does not merge anything itself: `gh pr merge --auto` hands the PR to
+# GitHub, which waits for the required `lint` check exactly like an agent PR
+# does. Branch protection stays the gate; this only removes the human click.
+#
+# Why a workflow at all: there is no native "auto-merge Dependabot" switch —
+# dependabot.yml has no such key, and the per-PR auto-merge button needs a human.
+#
+# `on: pull_request` (never pull_request_target): Dependabot-triggered runs get a
+# read-only GITHUB_TOKEN by default but have respected the `permissions:` key
+# since Oct 2021, so plain pull_request suffices. pull_request_target would run
+# with a write token in base-branch context — unnecessary risk for a job that
+# never checks out the PR's code.
+on: pull_request
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  auto-merge:
+    if: github.actor == 'dependabot[bot]'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: dependabot/fetch-metadata@v2
+        id: meta
+        with:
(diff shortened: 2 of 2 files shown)


alxjrvs (commented):
**Scope reviewed:** commit `9c546d9` (new `.github/dependabot.yml` + `.github/workflows/dependabot-auto-merge.yml`) plus the uncommitted doc changes in `dot-claude/CLAUDE.md`, `dot-claude/DECISIONS.md`, and `dot-claude/skills/agent-friendly-repo/SKILL.md`. The `gh-stack` hunks in `git diff main...HEAD` are already on `origin/main` (local `main` is stale), so I excluded them.

**Findings (3):**

- `/Users/jarvis/Code/DevEnv/dotFiles/.claude/worktrees/dependabot-automerge/.github/dependabot.yml:22` — All three action pins are bare major tags (`actions/checkout@v4`, `jdx/mise-action@v2`, `dependabot/fetch-metadata@v2`), so Dependabot will only ever open **major** PRs; those are excluded from the minor/patch group, the workflow's allowlist `if` never evaluates true, and the auto-merge path is permanently inert. The change's actual net effect is a new stream of manual major PRs, the opposite of what the commit message, the file header, and DECISIONS.md claim. Fix: pin to full semver (or SHA + version comment) so minor/patch bumps exist to group.
- `/Users/jarvis/Code/DevEnv/dotFiles/.claude/worktrees/dependabot-automerge/.github/workflows/dependabot-auto-merge.yml:39` — `gh pr merge --auto` errors ("Pull request is in clean status") if the PR is already mergeable when it runs; if `lint` wins the race, the job goes red and the PR stays open needing a human click. A `|| gh pr merge --squash "$PR_URL"` fallback closes it. (Plausible, not confirmed — `lint` is slow, so the race is unlikely but real.)
- `/Users/jarvis/Code/DevEnv/dotFiles/.claude/worktrees/dependabot-automerge/dot-claude/skills/agent-friendly-repo/SKILL.md:196` — New step 9 added, old step 9 renumbered to 10, but step 10's report checklist wasn't extended to cover the Dependabot outcome, so an agent running the skill won't report whether it wrote `.github/` or deferred.

**Verified as correct (no findings):** `on: pull_request` rather than `pull_request_target` (fork PRs get a read-only token regardless of the `permissions:` block, and the job is actor-gated anyway); `$PR_URL` passed via `env` rather than direct interpolation, so no script injection; allowlist rather than denylist on `update-type`; `dependabot.yml` group schema (`update-types: ["minor","patch"]` is the correct vocabulary, not the `version-update:semver-*` form the workflow uses); repo has `allow_auto_merge: true`, squash-only, and no merge queue, so the workflow's stated preconditions hold today; `lint.yml` is secret-free, so the Dependabot-secrets caveat genuinely doesn't bite here.

**Out-of-scope observation:** the repo is still on **classic** branch protection (`strict: true`, contexts `["lint"]`, `enforce_admins: true`) with zero rulesets — the reverse of what the skill and `~/.claude/CLAUDE.md` prescribe as the single source of truth. Not introduced by this PR, but the skill's own checklist would flag it.

PR #101, opened 2026-08-04
docs(agent-friendly-repo): encode the Dependabot auto-merge recipe

alxjrvs:
Follow-up to #97, which landed the mechanism on this repo (`.github/dependabot.yml` + the auto-merge workflow). This is the reusable half, so the next repo gets it from the skill instead of a re-derivation.

## What

- **`agent-friendly-repo` skill** — a "Dependabot auto-merge" checklist block (the workflow shape plus the four constraints), procedure step 9, and a guardrail.
- **`dot-claude/CLAUDE.md`** — the rule, in the repo merge & branch-protection defaults section.
- **`dot-claude/DECISIONS.md`** — the why, under Branch protection.

## The load-bearing points

Every one of these fails *quietly* rather than loudly, which is why they're worth writing down:

| | |
|---|---|
| The checklist already unblocks it | No required human review is the precondition. There is simply no native switch — `dependabot.yml` has no automerge key, and the per-PR button needs a human |
| `on: pull_request`, never `pull_request_target` | Dependabot runs get a read-only token by default but have respected `permissions:` since Oct 2021 — no reason to hand a write token to a base-context run |
| Actions secrets unavailable to Dependabot runs | A non-hermetic CI never auto-merges and never says why |
| `GITHUB_TOKEN` can't add a PR to a merge queue | This and the skill's optional merge-queue step are mutually exclusive without a PAT/App token |
| Allowlist, never `!= major` | An empty `update-type` has to fail closed to a manual PR |

## Verification

Docs only — no executable change. `biome check` clean, guard regression suite 33/33, gitleaks clean.

Worth noting: #97 merged mid-flight carrying only the two `.github` files, because a `git rebase` autostash reapplied these three doc edits without the index and the commit captured only what was staged. The split is harmless — #97 is functional standalone — but it's the reason this is two PRs rather than one.

https://claude.ai/code/session_018TCXxDahKWEKqsAnuTCKTr

### DIFF
--- a/dot-claude/CLAUDE.md
+++ b/dot-claude/CLAUDE.md
@@ -241,6 +241,13 @@ the recipes below are the reference it encodes.
   disable+re-enable the ruleset rather than adding a standing bypass. Find real check names via
   `gh api repos/<owner>/<repo>/commits/<branch>/check-runs --jq '.check_runs[].name'`.
   - The classic-protection fallback, for repos that cannot use rulesets, is in `DECISIONS.md`.
+- **Dependabot auto-merge** (opt-in, per repo): no required human review already unblocks it, but
+  nothing fires `--auto` for the bot, so it takes a one-job workflow
+  (`.github/workflows/dependabot-auto-merge.yml` here is the reference). `on: pull_request` with
+  an explicit `permissions:` block — never `pull_request_target`. Gate on an **allowlist**
+  (`update-type == minor || == patch`), never `!= major`. It cannot coexist with a merge queue
+  (`GITHUB_TOKEN` can't add a PR to one), and it silently never fires if CI needs Actions secrets
+  (Dependabot runs get only *Dependabot* secrets). Details in `DECISIONS.md`.
 
 ## Agent secret access
 
--- a/dot-claude/DECISIONS.md
+++ b/dot-claude/DECISIONS.md
@@ -315,6 +315,34 @@ EOF
 Prefer a ruleset. GitHub takes the *union* of classic protection and rulesets, so running both is a
 footgun: you edit one and the other silently still applies.
 
+### Dependabot auto-merge: a workflow, because there is no switch (2026-08-03)
+
+The agent-friendly checklist already unblocks it — **no required human review** means a Dependabot
+PR needs only a green aggregate check — but nothing fires `--auto` on the bot's behalf:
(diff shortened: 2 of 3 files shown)


alxjrvs (commented):
Reviewed the diff (dependabot config + auto-merge workflow, boomfile `gh extensions` section, and the doc/skill updates). Verified the runtime facts against the live repo (`gh api`: `allow_auto_merge: true`, classic protection with required context `["lint"]`, no ruleset, no merge queue) and against the Dependabot PRs that already landed.

Findings:

- `.github/dependabot.yml:7` — Every action is pinned to a floating major tag (`checkout@v4`, `mise-action@v2`, `fetch-metadata@v2`), so Dependabot only ever proposes **major** bumps, which this config deliberately excludes from the minor/patch group. Confirmed empirically: the first Dependabot run opened PRs #98/#99/#100, all majors, all individual, none grouped, and the allowlist step was skipped on each (`autoMergeRequest: null` on #98). The header comment's "minor+patch land grouped in one weekly PR, which … auto-merges once `lint` is green" describes behavior that cannot occur on this repo today — the auto-merge machinery is inert until the pins become exact versions.
- `.github/workflows/dependabot-auto-merge.yml:39` — `gh pr merge --auto` errors with "Pull request is in clean status" when nothing is left blocking the PR. On a `reopened` event (or a manual workflow re-run) after `lint` is already green, the step exits non-zero: red `auto-merge` check, no auto-merge enabled, manual click still needed. No fallback or `continue-on-error`.

Things I checked and found sound: the `if: github.actor == 'dependabot[bot]'` gate plus workflow-level `permissions:` is GitHub's own documented pattern and is safe under `on: pull_request` (fork PRs get a read-only token regardless); the allowlist (not `!= major`) fails closed as claimed and demonstrably held on the three major PRs; `lint.yml` is genuinely hermetic (no Actions secrets) and passed on all three Dependabot PRs, so the "secrets break it" precondition is satisfied; `allow_auto_merge` is on and there is no merge queue, so neither documented blocker applies; the `groups` schema uses the correct short `update-types` values; and the `gh extensions` boomfile section is correctly ordered after `packages` with `gh` supplied by `mise.toml`, matching the existing `lefthook install` / `tldr --update` precedent for PATH availability.

PR #102, opened 2026-08-05
docs(stacks): auto-merge can't land a stack — fix the claim, make ship stack-aware

alxjrvs:
Follow-up correction to #96.

## The error

#96 adopted `github/gh-stack` and recorded that `allow_auto_merge=true` was *"compatible: auto-merge coexists with stacks"*, inferred from `gh stack unstack` leaving a queued/auto-merge-enabled PR stacked. That sentence describes GitHub refusing to **unstack** a PR with a pending merge intent — it is not a completion path.

GitHub is explicit:

> Auto-merge is not supported for stacked pull requests.

> Merging a stacked pull request requires the Stacks API. The legacy pull request merge endpoints can't merge a stack.

...which is precisely the endpoint `gh pr merge` calls.

## Why it mattered

`--auto` is the entire reason the agent completion path is unattended: **it waits for green.** `gh stack merge` does not — it checks only that each PR is open and non-draft, then asks GitHub to merge *now*, so a pending or red aggregate check fails the whole all-or-nothing batch.

So "stacked PRs are the standing preference for agent work" and "agents complete work unattended" were quietly in conflict on every repo without a merge queue: an agent could build and submit a stack it had no way to land.

## Changes

- **`agent-friendly-repo`** — correct the claim. Promote the merge queue from *optional* to **required for unattended stacks**: with a queue, `gh stack merge` enqueues and the queue lands the stack once checks pass, which is the only true `--auto` analogue. Also records that checks are enforced **per PR against the stack's base branch** (so the default-branch ruleset governs every layer — no unguarded intermediate PR, but a required human review blocks all of them), and drops the stale "queue support still rolling out" hedge — it is complete, with sizing caveats instead.
- **`ship`** — was entirely stack-blind, the sharpest edge of the same error: a stacked PR's base is the branch below it, so the skill's documented final step `gh pr merge --auto --squash` would have merged a layer **into its parent branch** rather than `main`, collapsing the stack. Now runs `gh stack view` first and branches to a `sync` → `submit` → `merge` pipeline; on a queue-less repo it submits and hands the stack back rather than merging speculatively.
- **`CLAUDE.md` / `DECISIONS.md`** — the rules and the why.

## Deliberately not done

- **No `boomfile.toml` change.** The `gh extensions` section from #96 already installs `github/gh-stack`; nothing was missing there.
- **`Bash(gh stack merge:*)` NOT added to `permissions.allow`.** By the logic of the existing `Bash(gh pr merge:*)` rule it arguably belongs, but it is a different command with a different risk shape — queue-less, it merges N PRs immediately instead of waiting for green — and `CLAUDE.md`'s own rule is that settings get asked about, not added in passing. Left as an open question with the gap written into the contract, so an agent hits documented behavior rather than a silent denial.
- **No repo settings mutated.** Enabling a merge queue on this repo is an `agent-friendly-repo` action requiring explicit confirmation.

## Verification

Docs-only (4 markdown files). `gitleaks detect` clean; guard regression suite 33/33.

### DIFF
--- a/dot-claude/CLAUDE.md
+++ b/dot-claude/CLAUDE.md
@@ -89,6 +89,10 @@ Rules that apply whatever you are touching:
   classifier's own prescribed mechanism for letting background/unattended jobs land a finished
   PR. It authorizes `gh pr merge --auto` — GitHub's gated queue, which still waits on required
   checks — and does **not** loosen the "no direct `git` merge/push onto `main`" rule.
+  **It does not cover `gh stack merge`**, which is a different command and so an unenumerated
+  decision, not an oversight to be quietly fixed: unlike `--auto`, a queue-less `gh stack merge`
+  merges immediately rather than waiting for green. Adding `Bash(gh stack merge:*)` is the
+  open question; until it is answered, an unattended job submits a stack and hands it back.
 - **Worktree-checkout guard** — `PreToolUse` hook (matcher `Bash` →
   `~/.claude/hooks/worktree-checkout-guard.sh`, ordered *before* rebase-guard). Denies a
   `git checkout/switch <branch>` that would fail with "'<branch>' is already used by worktree".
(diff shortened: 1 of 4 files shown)


alxjrvs (commented):
Reviewed the 4-commit diff vs `main` (2 new YAML files, 1 new boomfile `run` step, and a large docs/skill surface). Findings:

- `boomfile.toml:322` — The `gh extensions` run step calls `gh` with no `command -v gh` guard. On the documented fresh-machine bootstrap, mise-installed `gh` is exposed only by `eval "$(mise activate zsh)"` in the interactive rc, which boom's already-running process never evaluated, so the step exits 127 and (by design, no `|| true`) fails the sync. Section ordering fixes install order, not inherited PATH; the sibling Claude-CLI step at line 218 *does* guard. Masked here only by an undeclared brew `gh` at `/opt/homebrew/bin/gh`.
- `.github/dependabot.yml:22` — All three action refs are floating major tags (`@v4`, `@v2`, `@v2`), so Dependabot only ever proposes majors, which the group excludes; the minor/patch group and the entire `dependabot-auto-merge.yml` workflow are inert as configured.
- `dot-claude/skills/ship/SKILL.md:68` — Guardrail "stop after step 5" now collides with the new stack path, whose step 5 *is* `gh stack merge`; a "don't auto-merge" instruction on a stacked branch reads as "merge the stack", contradicting the stack section's own "stop after step 4".
- `boomfile.toml:322` — `gh extension list` requires auth (verified: exits 4, empty stdout, even with the extension installed), so install-if-absent never matches on an unauthenticated machine and re-installs on every sync. Non-fatal (repeat install warns and exits 0) but the guard doesn't do its job.
- `dot-claude/skills/ship/SKILL.md:51` — `gh stack sync` fast-forwards trunk, and agents always run in linked worktrees while the primary checkout holds `main` — the repo's own documented "branch 'main' is already used by worktree" collision class, unmentioned in the new guidance, and invisible to the worktree-checkout guard.

Verified as fine along the way: boomfile section ordering (`gh extensions` sits between `packages` and `macOS defaults`), `on = "sync"` matching every other run step, dependabot `groups`/`update-types` schema, the allowlist-not-denylist gate, `on: pull_request` + explicit `permissions:` for the Dependabot token, `lint.yml` being hermetic (no Actions secrets), no merge queue on the repo (so no `GITHUB_TOKEN`-enqueue conflict), and no stale step-number cross-references in `agent-friendly-repo/SKILL.md` after the renumber.

PR #103, opened 2026-08-05
fix(stacks): close the four gaps that left the stacked-PR path unexecutable

alxjrvs:
Follow-up to #102. That PR documented how stacks work; this one makes them actually run.

Each gap was found by asking of every existing mechanism *"does `gh stack` match this?"* — the answer was **no** every time, and every no was silent.

## 1. `Bash(gh stack merge:*)` added to `permissions.allow`

#102 deferred this as an open question. It shouldn't have: `gh pr merge` **cannot merge a stack at all**, so an unattended job had no way to land one — the preferred shape for agent work was unreachable in exactly the mode that matters.

It's narrower than it looks. Stack merges cannot bypass merge requirements, so under the `agent-friendly-repo` ruleset it can only land a stack GitHub already considers mergeable. The real caveat is **timing, not privilege**: queue-less it merges *now* rather than waiting for green. That restraint lives in `ship` as prose, and prose is advisory — stated plainly rather than pretended away.

## 2. The PR-review hook never fired on a stack

Its trigger arms were `gh pr create` and `git push`. **`gh stack submit` matches neither** — it creates PRs through the Stacks API and pushes *inside the gh process*, so no `git push` Bash call ever reaches `PostToolUse`.

The entire argument for that hook was that 1,113 org PRs went unreviewed. Adopting stacks would have routed precisely the **largest** changes — the ones stacking exists for — around the reviewer.

Fixed in both places it has to be (the settings.json `if` handler *and* the script's own belt-and-braces `case` gate). Two honest limits, both documented in-line:
- Coverage is **partial**: it resolves the PR for the checked-out branch, so a submit reviews one layer, not the stack. Reviewing all N would mean N detached `claude -p` runs per submit.
- **`gh stack sync` is deliberately excluded.** It also force-pushes every branch, but a cascade-rebase replays the same content under new SHAs — firing on it means a full review per layer every time trunk moves. Cost without signal, and the per-SHA lock can't dedupe it because the rebase is what changes the SHA.

## 3. `rebase-prs` hand-rolled the cascade `gh stack sync` does natively

It instructed the agent to `git switch` each branch and rebase onto its parent — strictly worse than the tool: it force-pushes branch by branch, so a failure midway leaves the stack half-rebased with children pointing at commits that no longer exist, and it can't reconcile the stack object on GitHub. Straight *native over special*. Also resolves the oddity that the skill which motivated adopting `gh stack` was still competing with it.

## 4. Three `gh` extensions were installed by hand and undeclared

`dlvhdr/gh-dash`, `meiji163/gh-notify`, `actions/gh-actions-cache` — so **every fresh machine came up without them**. Exactly the drift this repo exists to prevent, and invisible because the `gh extensions` section existed and looked complete. Declared rather than uninstalled, since each has a live consumer.

## Checked, not assumed — and therefore *not* changed

`gh stack` is at v0.1.0, and v0.1.0 **is** the latest release (published 2026-07-29). The documented install-only/no-upgrade trap isn't biting yet. #96 said to revisit "if the version actually drifts far enough to bite" — it hasn't, so an upgrade step or version-drift check would be machinery for a hypothetical. Re-check when v0.2 lands.

## Verification

shellcheck + shfmt clean · biome clean · `jq` valid + all four settings.json content guardrails pass · taplo clean · gitleaks clean · **33/33** guard regression tests · new trigger gate tested against 7 command shapes (fires on `gh stack submit`/`gh pr create`/`git push`, skips `gh stack view`/`sync`/unrelated).

## Still open — your call, not mine

`dotFiles` runs **classic** branch protection (`lint`, strict) with **zero rulesets**, and has no merge queue. So stacks work here but only *supervised*. Enabling a queue would make them unattended — at the cost of the Dependabot auto-merge workflow from #97, since `GITHUB_TOKEN` can't enqueue. I did not touch repo settings; that's an `agent-friendly-repo` action needing explicit confirmation.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -53,7 +53,7 @@ The imperative residue the config can't express. A boom hook is a `hooks/<name>.
 Also NOT boom hooks — Claude Code hooks, linked into `~/.claude/hooks/`:
 
 - **`rebase-guard.sh` / `worktree-checkout-guard.sh`** — `PreToolUse` guards (see `dot-claude/CLAUDE.md`). They are the only deterministic enforcement in the setup.
-- **`pr-review.sh`** — `PostToolUse`: runs the adversarial review locally after `gh pr create` / `git push` and posts it as a PR review + a `claude-review` commit status. Backgrounds itself so it can never block a turn.
+- **`pr-review.sh`** — `PostToolUse`: runs the adversarial review locally after `gh pr create` / `git push` / `gh stack submit` and posts it as a PR review + a `claude-review` commit status. Backgrounds itself so it can never block a turn. The `gh stack submit` arm matches neither of the others (Stacks API + in-process push, so no `git push` Bash call), and covers the checked-out layer only.
 - **`hooks/tests/`** — `run.sh` + `cases.tsv`, a hermetic regression suite for the two guards (throwaway git fixtures in `$TMPDIR`, no network, <2s). Wired into `lint.yml` and lefthook pre-commit. **Add a case before changing a guard**: these are 200+ lines of security-relevant shell, and shipping them untested is how a `--dry-run` substring in an unrelated commit message came to disable the no-push-to-main rule.
 
 Small steps (`chmod 700 ~/.ssh`, `lefthook install`) are inline `run` (`on = "apply"`) entries, not files.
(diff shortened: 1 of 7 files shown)


alxjrvs (commented):
Reviewed commit `84f054d` against `origin/main` (11 files: settings.json, pr-review.sh, boomfile.toml, rebase-prs skill, docs). Gates all pass locally: `shellcheck -x`, `shfmt -d -i 2 -ci -sr`, `taplo check --no-schema`, `biome check`, `jq -e`, and the 33-case guard suite. I also verified the gh-stack claims against the installed `gh stack` v0.1.0 help — `sync`/`rebase`/`merge` semantics, the atomic `--force-with-lease --atomic` push, and the "cannot bypass merge requirements" line are all accurate, and all three newly declared extensions match their owner-qualified greps in real `gh extension list` output.

Findings:

- `dot-claude/skills/rebase-prs/SKILL.md:12` — The stack gate has no working detection: `gh stack view` only reads per-clone `.git/gh-stack` (never inherited by a fresh agent worktree), and step 2's `gh pr list --json` omits `baseRefName`, so "peel stacked ones off" has nothing to test. Meanwhile step 3's old "for a stack, rebase onto the parent" fallback was deleted — an undetected stack now gets each layer rebased onto `origin/main` and force-pushed, duplicating parents' commits into children. Strictly worse than before the change. Adding `baseRefName` to the `gh pr list` fields (a PR whose base is another open PR's head) would give the section a real test.
- `dot-claude/hooks/pr-review.sh:54` — `gh stack push` is the stack analogue of `git push` (which *is* a trigger) and pushes genuinely new commits, but isn't in the new `case`; the comment only justifies excluding `sync`. Follow-up commits on an already-submitted stack get no review and no `claude-review` status on the new head SHA — fatal if that status is ever promoted to required.
- `dot-claude/hooks/pr-review.sh:112` — `claude -p "/code-review"` receives no PR base. This is the first time the hook fires on a branch whose base isn't the default branch; `/code-review` resolves `@{upstream}...HEAD` / `main...HEAD`, so a layer's review either covers every layer below it (duplicate findings, `claude-review=failure` on a layer whose own diff is clean) or is empty because submit just pushed the branch.
- `dot-claude/DECISIONS.md:381` — The new `###` heading was inserted at line 339, above the prior entry's closing "Deliberately not adopted in the same change: `gh skill install github/gh-stack`" paragraph, which now reads as part of the 2026-08-04 entry that enumerates "Four gaps" and installs no skill. Move the new entry after line 386.
- `dot-claude/settings.json:34` — `Bash(gh stack merge:*)` auto-approves a bare `gh stack merge`, which non-interactively uses the last-used merge method. The added docs assert the only caveat is "timing, not privilege"; merge method is a second one. On a repo not yet aligned by `agent-friendly-repo`, that lands merge commits against the squash-only doctrine, enforced only by `ship`'s prose.

PR #104, opened 2026-08-05
chore(actions): bump jdx/mise-action v2 → v4 (supersedes #99)

alxjrvs:
Supersedes #99. Same upgrade, correct pin.

## Why not just merge #99

Dependabot proposed `jdx/mise-action@v4.2.3` — an exact patch — which is wrong on two counts:

- **It breaks this repo's convention.** Every other action here pins a floating major (`actions/checkout@v7`, `dependabot/fetch-metadata@v3`), so patch releases land silently and Dependabot only opens a PR on a major. An exact pin turns every future mise-action patch into a PR to triage.
- **It was already stale on arrival.** `v4.2.4` shipped 2026-08-01 — before the PR was reviewed — so merging as-proposed would have immediately queued another bump.

jdx maintains a real floating `v4` tag, and it points at `7e36c90`, the **same SHA as v4.2.4**. So `@v4` is simultaneously current and self-updating within the major.

## Verification

The upgrade itself is proven by #99's own CI: that PR changed `lint.yml`, so its green `lint` run executed with mise-action v4 and successfully ran `mise install shellcheck shfmt gitleaks jq taplo biome`. This PR points at the same code by a different tag.

Rebased onto current `main`, so it also carries `actions/checkout@v7` from the just-merged #98.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -21,7 +21,7 @@ jobs:
 
       # Set mise up (PATH + shims) but skip the full `mise install` — we only
       # need the lint tools, not the whole toolchain (node, bun, neovim…).
-      - uses: jdx/mise-action@v4.2.3
+      - uses: jdx/mise-action@v4
         with:
           install: false
       - name: Install lint tools
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -21,7 +21,7 @@ jobs:
 
       # Set mise up (PATH + shims) but skip the full `mise install` — we only
       # need the lint tools, not the whole toolchain (node, bun, neovim…).
-      - uses: jdx/mise-action@v2
+      - uses: jdx/mise-action@v4.2.3
         with:
           install: false
       - name: Install lint tools


alxjrvs (commented):
## Review complete — no findings

**Scope.** The branch (`mise-pin-fix`, PR #104) is fully pushed, so `git diff @{upstream}...HEAD` is empty; the working tree is clean. The real review scope is `git diff origin/main...HEAD`, which is a **single line** in `/Users/jarvis/Code/DevEnv/dotFiles/.claude/worktrees/gh-stack-support/.github/workflows/lint.yml:24`:

```
- uses: jdx/mise-action@v2   →   - uses: jdx/mise-action@v4
```

(Two commits — Dependabot's `@v2 → @v4.2.3`, then a re-pin to the floating `@v4` — net out to that one line.)

**What I verified against the upstream action, not just the diff:**

- The floating `v4` tag exists and resolves to `7e36c90` — identical to `v4.2.4`. jdx maintains floating majors consistently (`v1`, `v2`, `v3`, `v4` all present), so the pin won't silently go stale within the major, and the commit message's central claim checks out.
- `v4.0.0`'s only breaking change is the Node 20 → Node 24 runtime; `ubuntu-latest` supports it.
- The `v2 → v4` span carries three behavior changes that *could* have bitten this workflow, and none do: `v3.0.0`'s `mise.toml` env-var export is a no-op (the repo's `mise.toml` has no `[env]`), `v4.1.0`'s auto-`--locked` is gated behind `install:` (false here) and there is no `mise.lock`, and `v4.2.1`/`v4.2.3`'s PATH-export rework is the *fixed* state — `v4` is past the `v4.2.1` regression that would have dropped shims from `PATH`.
- `add_shims_to_path` still defaults to `true`, so `install: false` followed by an explicit `mise install shellcheck shfmt gitleaks jq taplo biome` still resolves for later steps. The actual run log confirms `Adding /home/runner/.local/share/mise/shims to PATH`, and `lint` is green on the exact HEAD SHA with no warnings. Non-preinstalled tools (`shfmt`, `taplo`, `biome`, `gitleaks`) passing rules out a silent fallback to the runner's system binaries.
- Consistent with `.github/dependabot.yml`: a major-only tag means Dependabot stays quiet on patches and only files majors, which are excluded from the auto-merge group and stay manual — exactly the documented intent.

**One non-blocking nit, not filed as a finding** (no runtime failure mode): the commit message for `03a0669` cites `dependabot/fetch-metadata@v3` as evidence of the repo's floating-major convention, but the tree still has `@v2` at `.github/workflows/dependabot-auto-merge.yml:27` — the v3 bump is open and unmerged as PR #100. The convention claim is right; that one example is just ahead of the tree.

PR #106, opened 2026-08-05
docs(stacks): make stacks the default shape and decline the merge queue

alxjrvs:
Owner's call, and it **reverses #103's conclusion.** #102/#103 argued the merge queue "stopped being optional" because `--auto` cannot land a stack and `gh stack merge` doesn't wait for green. The facts are unchanged; the *decision* is that fire-and-forget was never worth its price here.

## What the queue actually costs

- **Mutually exclusive with Dependabot auto-merge.** `GITHUB_TOKEN` cannot add a PR to a merge queue, so enabling one silently breaks #97 — trading a *daily* automation for an *occasional* one. Wrong way round.
- **Carries the `merge_group:` sequencing hazard** — enable it before CI reports on the queue's temp branches and every PR hangs forever. A permanent foot-gun in exchange for convenience on multi-layer changes.
- **Degrades the guarantee stacks exist for** — a queued stack may be split across consecutive merge groups, so all-or-nothing weakens to per-group.

## What "no queue" costs, honestly

The agent watches every layer to green before merging. #103 called that *"a babysitting loop, not a completion path"* — that was overstated, and this PR retracts it. `gh pr checks --watch` is a bounded, supported wait. It only degenerates if `main` moves faster than the stack can settle, which is a many-contributor problem these repos don't have. Hence the explicit rule: **sync, retry once, then report** — never loop indefinitely.

## Changes

- **`ship`** — the stack path is now a real completion sequence (watch every layer → `gh stack merge --yes --squash`) instead of a hand-back. Documents *why* every layer must be watched (checks are enforced per-PR against the stack's base at merge time, so a green bottom PR proves nothing), the strict-policy re-sync retry, and that `--squash` yields one commit per layer — which is what preserves the layering in linear history.
- **`agent-friendly-repo`** — queue demoted to skip-by-default, gated behind an explicit ask *and* a repo with no Dependabot auto-merge to lose. Removes the "unattended stacks require a queue" claim throughout.
- **`CLAUDE.md`** — completion now has two documented shapes rather than one plus a caveat; the standing preference names stacks as the default and no-queue as deliberate.

## The half that tooling can't do

Leaning in isn't just having `gh stack` installed — it's **deciding the layers before writing the code**, because splitting one large commit afterwards is archaeology. `CLAUDE.md` now carries the test:

> a layer is something that could be reviewed, **and reverted**, on its own — an enabling refactor, a schema change ahead of its consumers, a mechanical rename, docs separated from code. Not a split by file, by commit count, or to hit a size target.

With the counterweight attached, so this doesn't produce silly 1-line stacks: **don't stack a single reviewable change** (a one-layer stack is a PR with extra ceremony), and don't manufacture layers to satisfy the rule.

## Eating it

**This PR is deliberately not a stack**, under that same rule — it's one reviewable change to one doctrine, and splitting it would be exactly the manufactured layering the rule warns against.

`DECISIONS.md` also records the uncomfortable evidence: #102 and #103 were *themselves* stack-shaped (#103 built directly on #102's conclusions) and were shipped serially anyway. That's the clearest measure available of the gap between writing the preference down and reaching for it.

## Verification

gitleaks clean · 33/33 guard tests · biome clean · grepped `ship` and `CLAUDE.md` for leftover contradictory merge-queue guidance (the two remaining mentions are the skill pointer and the Dependabot incompatibility note, both now consistent).

### DIFF
--- a/dot-claude/CLAUDE.md
+++ b/dot-claude/CLAUDE.md
@@ -164,12 +164,25 @@ Claude Code isolates background/subagent work into git worktrees by default
 overridden). Each agent gets `.claude/worktrees/<name>` on a **freshly created branch** off
 `origin/<default-branch>`.
 
-- **Prefer a stack of small PRs over one large one**, via the **official** `github/gh-stack`
-  extension (`gh stack`, public preview since 2026-07-30) — installed by the dotFiles boomfile's
-  `gh extensions` section, never a same-named community fork. Reach for it when a change is
-  bigger than one reviewable PR: `gh stack init` → `gh stack add` per layer → `gh stack submit`
-  (one PR per branch, each based on the one below) → `gh stack sync` after a layer lands. It owns
-  the cascading rebase across the whole stack, which is the part that is miserable by hand.
+- **A stack of small PRs is the default shape for multi-part work**, via the **official**
+  `github/gh-stack` extension (`gh stack`, public preview since 2026-07-30) — installed by the
+  dotFiles boomfile's `gh extensions` section, never a same-named community fork.
+  `gh stack init` → `gh stack add` per layer → `gh stack submit` (one PR per branch, each based
+  on the one below) → `gh stack sync` after a layer lands. It owns the cascading rebase across
+  the whole stack, which is the part that is miserable by hand.
+  - **Decide the shape before writing code, not at ship time.** Once the work is one large
+    commit, splitting it is archaeology. If the plan has separable layers, cut the branches up
+    front — that is the entire ergonomic win, and it is unavailable retroactively.
+  - **What earns its own layer:** a change that could be reviewed, and reverted, on its own. In
+    practice that means a refactor that enables the feature (not the feature), a schema or type
+    change ahead of its consumers, a mechanical rename or move separated from behavior, or docs
+    separated from code. **What does not:** splitting by file, by commit count, or to hit a size
+    target. A layer nobody could review in isolation is not a layer.
+  - **Don't stack a single reviewable change.** A one-layer stack is a PR with extra ceremony.
+    The test is whether a reviewer would want the layers separately, not whether the diff is big.
+  - **Stacks are the default, not a mandate.** When work genuinely resists layering — a change
+    that is atomic by nature, or an urgent fix — ship one PR and say why in a sentence. Don't
+    manufacture layers to satisfy the rule.
   - `gh stack submit`/`push`/`sync` are **not** matched by the rebase-guard (which tokenizes for
     `git push` / `gh pr create` specifically), so the guard is silent on them. That is not
     permission to publish a stale stack — `gh stack sync` (fetch + cascading rebase + push) is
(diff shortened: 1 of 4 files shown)


alxjrvs (commented):
Reviewed `git diff main...HEAD` (11 commits: gh-stack adoption, Dependabot auto-merge, biome trim, statusline casing, and the top commit that reverses the merge-queue decision). Working tree is clean, so nothing extra in scope. The executable surface is small (one shell hook, one TS hook, two workflows, a boomfile section); most of the diff is doctrine docs that agents execute, so I treated contradictions there as real defects.

Findings:

- `dot-claude/CLAUDE.md:97` — The `Bash(gh stack merge:*)` permission bullet still says an unattended job on a queue-less repo should "submit the stack and hand it back instead of merging speculatively" and credits that restraint to `ship`. The same commit rewrote `ship` and lines 197-221 to prescribe the opposite (watch every layer green → `gh stack merge --yes --squash`). Since every repo is now queue-less by decision, this stale bullet applies universally, and it sits earlier in the file than the corrected guidance.
- `dot-claude/skills/agent-friendly-repo/SKILL.md:200` — Step 6 still applies `strict_required_status_checks_policy: true` while step 8 now skips the merge queue by default. Line 144 of the same skill names the queue as the fix for strict's serialization churn, so the checklist now bakes in that churn with its stated remedy removed; for stacks it makes `ship`'s "main moved, merge rejected, re-sync and retry" the default path. Not revisited in DECISIONS.md.
- `dot-claude/skills/ship/SKILL.md:53` — The stack path says `gh stack submit` creates drafts non-interactively, then goes straight to `gh stack merge`, which the skill's own guardrail (line 80) says requires non-draft PRs. No `gh pr ready` step and no instruction to pass `--open` when the intent is to land.
- `hooks/claude_statusline.ts:20` — The `thegnarco` → `TheGnarCo` casing change flows into `pointsElsewhere()`, which substring-matches `.git/config` case-sensitively. Existing checkouts (lowercase URL) now report false drift on every `boom verify` and hit the `rm -rf` + re-clone branch on the next `boom source`, for the same repo.
- `dot-claude/hooks/pr-review.sh:82` — PLAUSIBLE. The hook runs after the push has already landed, then invokes `/code-review`, whose scope starts at `git diff @{upstream}...HEAD` — empty once HEAD equals upstream, with `git diff HEAD` (clean tree) as the equally-empty fallback. Result is a green `claude-review` status on code never examined. Pre-existing for the `git push` arm; the new `gh stack submit` arm inherits it and the new comment asserts per-layer coverage that doesn't hold.
- `dot-claude/skills/agent-friendly-repo/SKILL.md:109` — Recipe pins `dependabot/fetch-metadata@v2` while this repo's own reference workflow is `@v3` (bumped in d7645df); a repo configured from the skill immediately gets a major-bump PR the allowlist won't auto-merge.

Things I checked and found sound: the boomfile `gh extensions` section ordering (`gh` does come from `mise.toml:32`, and the section follows `packages`), `on = "sync"` matching every other run step in the file, the owner-qualified greps, the `biome.json` default-trimming (2.5.3 defaults confirm formatter/linter enabled and indentWidth 2), the Dependabot allowlist gating and `on: pull_request` + explicit `permissions` choice, and the `ship` single-PR step numbering ("stop after step 5" correctly precedes the auto-merge step).

PR #111, opened 2026-08-06
docs(secrets): correct seven drifted claims and fold in 2026 vendor facts

alxjrvs:
The audit verified these against the code and the machine. Each was
accurate-sounding and wrong, and several were load-bearing.

Understated risk:
- "neither the SA token nor the resolved secret reaches a Bash subprocess,
  the transcript, or OTEL" — true of the SA token, false of resolved
  secrets. op-agent prints results to stdout by design (that is how a
  resolver returns a value) and Bash stdout is model context. Nothing
  structural prevents it; the deny list does.
- "That vault is the entire blast radius" — that is the SERVICE ACCOUNT's
  radius. The agent runs as you, so `op vault list` from an ordinary Bash
  call enumerates the whole account. Cross-vault reads are prompt-gated, so
  an unattended job can't reach them; the exposure is an attended session
  and reflexive approval. Gate scope was not characterized — desktop
  integration typically authorizes per-session, not per-item.
- "the PAT lives only in 1Password — no keychain cache" — true of storage,
  not residency: `cache --timeout=900` holds it in daemon memory for 15
  minutes, readable same-user via `git credential fill`.
- `Read(~/.ssh/id_*)` reads as protecting those files; deny is tool-scoped
  and governs the Read tool only.
- "No wrapper, no exported env var" — gh-mcp-stdio exports the PAT into a
  long-lived process, because Claude Desktop supports nothing better.

Overstated risk (drift runs both ways):
- DECISIONS.md called the GitHub MCP user-scoped, "reachable from every
  session in every repo". It is project-scoped to ~/Code/SU-SRD. An
  overstated risk is not the safe kind of wrong — it misdirects attention,
  and it is how the "0 calls = unused" misdiagnosis happened before.

Neither accurate nor inaccurate, just absent:
- `op run --env-file` is called the canonical MCP pattern. It is vendor-
  endorsed (1Password's own Nov 2025 guidance prescribes it) and has zero
  instances here — every live consumer uses a path the docs frame as the
  exception. Marked as aspirational rather than descriptive.
- The `render` MCP server is a second undocumented consumer of `op-agent
  header`; neither it nor `github` is declared in the boomfile, so a fresh
  machine reproduces neither and the mcp-list check can't see an absent one.

Vendor facts (fetched, not recalled — several post-date training):
- excluded vaults also include your DEFAULT SHARED vault;
- SA token rotation is web-UI only and unscriptable, and cannot add an
  expiry to an existing token;
- the audit log, usage reports and Events API are BUSINESS-tier. On a
  personal plan an agent's `op read` leaves no retrievable record at all,
  so scope + expiry are the entire story. Worth confirming the tier.
- the daily rate limit is per ACCOUNT (1,000 on Individual/Family), so the
  credential cache is doing quota work, not just latency work;
- developer.1password.com now 301s to www.1password.dev.

Also records the postmortem lesson in DECISIONS.md: when an incident
produces a rule, check the rule landed in a layer that can enforce it.

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


alxjrvs (commented):
Reviewed the 5-commit branch (`main...HEAD`): deny-list rework + regression checks, SA token expiry, pr-review hardening, and a large docs correction pass. I read every hunk, opened `hooks/op-agent.sh`, `dot-claude/hooks/pr-review.sh`, `dot-claude/settings.json`, `boomfile.toml` and both enforcement points in full, and empirically verified the shell/regex behaviors below (`cut`, integer division, the grep/JS-regex guardrails).

Verified as working: the three new deny-floor assertions all match the real `dot-claude/settings.json` (grep `-F` form and the JS-regex-escaped boomfile form both tested true), shellcheck is clean on both scripts, and the `autoUpdate: true → false` flip is documented in `dot-claude/CLAUDE.md:84`.

Findings:

- `dot-claude/settings.json:39` — The rewrite dropped `Bash(~/.local/bin/op-agent secret:*)` with no path-qualified replacement, so the absolute-path spelling (which this same file uses three times, at lines 19/20/24) is no longer denied — `~/.local/bin/op-agent header op://claude-agent/claude-git-pat/credential` was blocked before and is not now.
- `hooks/op-agent.sh:87` — `cut -d. -f2` passes a delimiter-less line through whole, so the `[[ -n "$payload" ]]` guard never fires for a non-JWT token; the expiry warning then silently no-ops and is indistinguishable from "token is open-ended".
- `hooks/op-agent.sh:232` — Integer division truncates toward zero, so a token expired <24h ago gives `sa_days == 0`, skips the EXPIRED branch, and `op-agent status` returns 0 — `boom verify` stays green for a full day after the SA token has lapsed, contradicting the documented "fails once it has lapsed".
- `dot-claude/hooks/pr-review.sh:130` — `--allowedTools` is an additive allow list, not a ceiling; the detached reviewer inherits `defaultMode: auto` + `skipAutoPermissionPrompt`, so unlisted tools fall to the probabilistic classifier rather than being blocked. Needs `--permission-mode default` or `--disallowedTools` to make the "read-only, nothing else" claim true.

alxjrvs (commented):
Reviewed the 10-file diff (`main...op-docs-drift`). Verified claims empirically rather than by reading: dumped boom's embedded `check.ts` to confirm `present` patterns really are JS regexes (the escaped boomfile patterns are correct — no finding there), probed the permission matcher to confirm `Bash(op:*)` is token-aware and does *not* catch `openssl` (the docs' claim holds), unit-tested the JWT expiry decoder, and ran the repo's own shellcheck/shfmt/taplo gates (all clean).

Findings:

- `dot-claude/hooks/pr-review.sh:131` — `--allowedTools` is an additive allow-list, not a restriction. A spawned `claude -p ... --allowedTools "Read,Grep,Glob"` still ran a Bash command in my test; only `--disallowedTools "Bash"` removed the tool. The reviewer keeps Bash/Write/Edit/WebFetch/Task, so the "it just can't execute anymore" hardening does not hold — `curl`-based exfil from a poisoned diff still works, silently, since the block is `>/dev/null 2>&1 &`. (The specific `op-agent header` chain *is* closed, but by the new deny entries, not by this flag.)
- `lefthook.yml:56` (same shape at `.github/workflows/lint.yml:74`, `boomfile.toml:272`) — the new floor assertion is a substring grep, so moving `"Bash(op-agent:*)"` from `deny` into `allow` inverts the control while all three enforcement points stay green. `jq -e '.permissions.deny | index(...)'` would be array-aware; `jq` is already a dependency of that block.
- `hooks/op-agent.sh:232` — integer division truncates toward zero, so for the first 24h after the SA token expires `sa_days == 0`: the `-lt 0` fail branch is skipped, it prints "expires in 0d", returns 0, and `boom verify` stays green while every secret path is already dead (measured at -300s/-3600s/-43200s offsets).
- `.github/workflows/lint.yml:74` — the floor check asserts 3 of the 7 deny entries; `Bash(security find-generic-password:*)` and `Bash(~/.local/bin/op-agent:*)` can still be deleted with every gate green.
- `dot-claude/CLAUDE.md:165` (also `CLAUDE.md:56`, `dot-claude/DECISIONS.md:98`) — all three record the reviewer as read-only, so the durable postmortem asserts a control stronger than the code, which is the same failure mode this PR is fixing elsewhere.

PR #117, opened 2026-08-08
chore(settings): drop what equals the default, and correct three false claims

alxjrvs:
settings.json's own rule is that it carries "only deliberate divergences from
defaults", and the enumeration in dot-claude/CLAUDE.md is the contract. Four
entries broke that rule, checked against the published settings schema rather
than assumed:

  autoScrollEnabled: true    schema default is true
  statusLine.padding: 0      schema: "defaults to 0"
  voiceEnabled: true         schema: "Legacy alias for voice.enabled; prefer the
                             voice object" — set alongside `voice`, so this was
                             one setting written twice
  SPACEBASE_* (four vars)    consumer deleted in #113. The retained-because-the-
                             resolver-is-fiddly argument does not survive
                             contact with NINETY_API_TOKEN_COMMAND, which is the
                             identical shape two lines away. The vaulted
                             spacebase-api-key item is untouched.

Two that LOOK redundant and are not, so they stay: `theme: "auto"` (schema
default is "dark") and `inputNeededNotifEnabled: true` (default false). A review
agent recommended deleting both; the schema says otherwise.

teammateMode also equals the current default and is kept deliberately, now
documented as a defensive pin rather than a divergence: that default was `auto`
before v2.1.179, and a silent flip back would start opening split panes.

Doc corrections, all for claims that asserted a control stronger than the code —
the same failure mode this branch has been fixing everywhere:

  "The reviewer is read-only (--allowedTools)"  measured false; --allowedTools is
    additive pre-approval and confines nothing. Replaced with what actually holds
    (permissions.deny via --settings) and an instruction to re-run the probe when
    either file changes, because the previous text was an argument rather than a
    measurement.
  "boom verify fails when any server is down"  it matched only ✘, missing
    "! Needs authentication", and it cannot see the project-scoped github/render
    servers at all.
  "Advisory until a day-30 finding rate"  that clock restarts today: the
    numerator was produced by a grep that matched nothing, so 15 reviews carrying
    real findings all reported clean. Pre-2026-08-08 data cannot be used to
    judge this hook.

Co-Authored-By: alxjrvs <alxjrvs@gmail.com>
Claude-Session: https://claude.ai/code/session_017fqCjxH7JXUK3vA9aFD4FK

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


alxjrvs (commented):
I reviewed the full `main...HEAD` diff (17 files, ~1160 insertions), read the surrounding files, and **reproduced every guard finding through the repo's own harness** (`dot-claude/hooks/tests/run.sh` with ad-hoc case files). The shipped suite passes 71/71, including under real bash 3.2; `shfmt`, `shellcheck`, `biome`, `taplo`, `zsh -n` and the new plist grep all pass, and the lefthook `zsh-syntax` glob does match nested `zsh/*.zsh`.

Findings:

- `dot-claude/hooks/guard-lib.sh:87` — `_split` preserves newlines inside quotes/after `\`, but both guards transport segments through a newline-delimited `while read`, so the segment is re-split anyway. Reproduced: `git push \`⏎`origin main` is **allowed** (direct-push-to-main bypass), `git checkout \`⏎`feature-b` is **allowed**, and `gh pr create --body "line1`⏎`git push origin main was allowed"` is **denied** — the multi-line-message false positive the guard claims to have fixed.
- `dot-claude/hooks/guard-lib.sh:47` — the heredoc-delimiter regex matches `<<<` (which the comment says is "deliberately not matched") and a quoted `<<`, swallowing the rest of a multi-line command. Reproduced: `grep -q x <<< "text"`⏎`git push origin main` and the `git checkout feature-b` equivalent both bypass their guard.
- `dot-claude/hooks/pr-review.sh:188` — a missing/unparseable `CLAUDE-REVIEW-SUMMARY` trailer now `exit 0`s **before** `gh pr review`, so a review full of real findings is deleted and never posted. Previously the body was always posted.
- `dot-claude/hooks/pr-review.sh:189` — all four give-up paths post GitHub state `success` (green), contradicting the new `CLAUDE.md` claim that the status "fails CLOSED on an unparseable body"; if the context is ever promoted to required, every silent-failure path passes the gate.
- `dot-claude/hooks/rebase-guard.sh:191` — `--dry-run` changed from `allow` (exit) to `continue`, so a dry-run push on a behind branch is now denied, contradicting the adjacent comment and pinned by no test.

Non-findings worth knowing: the `/tmp/boom-verify.log` plist change is correct for launchd (no `~` expansion) though `/tmp` is periodically purged; `git checkout -b x && git checkout <held-branch>` short-circuits to `allow`, and a chain of relative `cd`s resolves only the last one — both pre-existing and fail-open.

PR #120, opened 2026-08-15
feat(skills): add the butter-stack skill so agents can copy the house shape

alxjrvs:
The same monorepo shape has been re-derived four times. `@RANDSUM`, `SU-SRD`, `OptFall` and `BinfiniteApp` each arrived at Bun workspaces + `apps/*`/`packages/*` + a strict-plus base tsconfig + Biome + a single `always()` CI gate **independently**, never templated from one another. That convergence is the argument for writing it down — until now it existed only as four separate `CLAUDE.md` files, so every new repo and every agent re-derived it from scratch.

Names it **the Butter Stack** (it goes on the Bun), with a backronym where every letter is a layer that was actually verified rather than filler:

| | | |
|---|---|---|
| **B** | Bun | runtime · pm · workspaces · tests · scripts |
| **U** | Unified workspace | `apps/*` + `packages/*`, nothing else |
| **T** | TypeScript | 7.0.2, strict-plus |
| **T** | TanStack | router · query · start |
| **E** | Edge-deployed | Netlify · Render · Convex |
| **R** | React | 19, behind a shared component library |

## Two parts carry more weight than the tool list

**The signature habit** — conventions get promoted from prose into executable `check:*` gates wired into the aggregate job (`check:tokens`, `check:hub`, `check:stories`, `validate:architecture`, `check:ci-aggregator`). A rule nothing enforces is a rule that has already drifted. This repo's own guard regression suite is the same instinct, which is why the skill states it as a rule rather than an observation.

**A deliberate-exceptions table.** Binfinite's vitest / TS 6.0.3 / `hoisted` linker, and randsum's off-catalog TS 6.0.3 pins, each look exactly like drift and are load-bearing. The randsum pins exist because `@astrojs/check` needs the TS6 compiler API, which TS7 deliberately does not ship — reasoning that until today lived only inside a closed issue. Recording them is the point: a tidying pass that "fixes" one of those breaks a build.

## Behaviour

- **Audit** is read-only and reports drift as a table.
- **Scaffold** writes only after an explicit go-ahead.
- Hands off to `agent-friendly-repo` for merge settings, since a `CI Success` gate is not real until it is the required check.

## Notes

`dot-claude/skills/*` is already glob-linked into `~/.claude/skills/` (boomfile line 226), so this needs **no boomfile change** — a new directory is picked up on the next `boom source`.

Local checks: `biome check` clean, `gitleaks` no leaks, `dot-claude/hooks/tests/run.sh` **71 passed**.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -141,6 +141,20 @@ Symlinked individually into `~/.claude/` (the `Claude` section of the boomfile):
 
 - `CLAUDE.md` — user-level global instructions (identity, preferences).
 - `settings.json` — **deliberately minimal**; only divergences from defaults (enumerated in `dot-claude/CLAUDE.md`). Don't add settings without asking.
+- `skills/*` — glob-linked into `~/.claude/skills/`, so a new directory is picked up with no boomfile change. Today: `agent-friendly-repo`, `ship`, `rebase-prs`, `butter-stack`.
+
+### The Butter Stack (`dot-claude/skills/butter-stack/`)
+
+**Butter, because it goes on the Bun** — the house shape for personal TypeScript work, and the one place it is written down so an agent can copy it rather than re-derive it per repo: **B**un · **U**nified workspace (`apps/*` + `packages/*`) · **T**ypeScript (strict-plus) · **T**anStack · **E**dge-deployed (Netlify/Render/Convex) · **R**eact.
+
+It is descriptive before it is prescriptive. `@RANDSUM`, `SU-SRD`, `OptFall` and `BinfiniteLLC/BinfiniteApp` arrived at it **independently**, never templated from one another — that convergence is the evidence, and it is why the skill records what four repos re-decided under pressure rather than a preferred toolchain.
+
+Two things in it matter more than the tool list:
+
+- **The signature habit** — conventions get promoted from prose into executable `check:*` gates wired into the aggregate CI job. A rule nothing enforces is a rule that has already drifted; this repo's own guard tests are the same instinct.
+- **The deliberate-exceptions table** — Binfinite's vitest / TS 6 / `hoisted` linker, and randsum's off-catalog TS 6.0.3 pins for `@astrojs/check`. Each looks like drift and is load-bearing. The table exists so a tidying pass doesn't remove them; **check it before calling anything drift**.
+
+The skill audits (read-only) or scaffolds (writes only after a go-ahead), and hands off to `agent-friendly-repo` for merge settings — a `CI Success` gate is only real once it is the required check.
 
 ## Guardrails
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -147,7 +147,7 @@ Symlinked individually into `~/.claude/` (the `Claude` section of the boomfile):
 
 **Butter, because it goes on the Bun** — the house shape for personal TypeScript work, and the one place it is written down so an agent can copy it rather than re-derive it per repo: **B**un · **U**nified workspace (`apps/*` + `packages/*`) · **T**ypeScript (strict-plus) · **T**anStack · **E**dge-deployed (Netlify/Render/Convex) · **R**eact.
 
-It is descriptive before it is prescriptive. `@RANDSUM`, `SU-SRD`, `OptFall` and `BinfiniteLLC/BinfiniteApp` arrived at it **independently**, never templated from one another — that convergence is the evidence, and it is why the skill records what four repos re-decided under pressure rather than a preferred toolchain.
+It is descriptive before it is prescriptive. `RANDSUM/randsum`, `SalvageUnion-io/SU-SRD`, `alxjrvs/optfall` and `BinfiniteLLC/binfinite-app` arrived at it **independently**, never templated from one another — that convergence is the evidence, and it is why the skill records what four repos re-decided under pressure rather than a preferred toolchain. (Those are the GitHub names; the local checkouts under `~/Code` are spelled differently, and `BinfiniteLLC/BinfiniteApp` is a directory name that does not resolve on GitHub.)
 
 Two things in it matter more than the tool list:
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -141,7 +141,7 @@ Symlinked individually into `~/.claude/` (the `Claude` section of the boomfile):
 
 - `CLAUDE.md` — user-level global instructions (identity, preferences).
 - `settings.json` — **deliberately minimal**; only divergences from defaults (enumerated in `dot-claude/CLAUDE.md`). Don't add settings without asking.
-- `skills/*` — glob-linked into `~/.claude/skills/`, so a new directory is picked up with no boomfile change. Today: `agent-friendly-repo`, `ship`, `rebase-prs`, `butter-stack`.
+- `skills/*` — glob-linked into `~/.claude/skills/`, so a new directory is picked up with no boomfile change. **Read the directory for the roster; it is deliberately not enumerated here** — a hand-maintained list of a glob-linked directory goes stale the next time a skill is added or removed, with no symptom, and nothing checks it.
 
 ### The Butter Stack (`dot-claude/skills/butter-stack/`)
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -152,7 +152,7 @@ It is descriptive before it is prescriptive. `RANDSUM/randsum`, `SalvageUnion-io
 Two things in it matter more than the tool list:
 
 - **The signature habit** — conventions get promoted from prose into executable `check:*` gates wired into the aggregate CI job. A rule nothing enforces is a rule that has already drifted; this repo's own guard tests are the same instinct.
-- **The deliberate-exceptions table** — Binfinite's vitest / TS 6 / `hoisted` linker, and randsum's off-catalog TS 6.0.3 pins for `@astrojs/check`. Each looks like drift and is load-bearing. The table exists so a tidying pass doesn't remove them; **check it before calling anything drift**.
+- **The deliberate-exceptions table** — a set of per-repo deviations that each look like drift and are load-bearing, so a tidying pass doesn't remove them. **Read it in the skill and check it before calling anything drift.** Deliberately not summarized here: copying those rows into always-loaded context is the same hand-maintained-duplicate problem as the skills roster above, and the table is the one thing in the skill that must never be consulted from a stale copy.
 
 The skill audits (read-only) or scaffolds (writes only after a go-ahead), and hands off to `agent-friendly-repo` for merge settings — a `CI Success` gate is only real once it is the required check.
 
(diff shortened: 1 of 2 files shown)


alxjrvs (commented):
Read the diff (doc-only: a new `dot-claude/skills/butter-stack/SKILL.md` plus a CLAUDE.md section) and checked it against the boomfile, the sibling skills, and the repo's own conventions. No text in the diff was addressed to a reviewing agent — the SKILL.md contains agent instructions by design, and I treated them as data, not as directions.

**Findings**

- **BLOCKING** — `dot-claude/skills/butter-stack/SKILL.md:48` — the required gate is specified as "one job with `if: always()` that `needs:` every other job" with no failure-propagation step, but an `always()` job whose steps never inspect `needs.*.result` **succeeds when its dependencies fail**, so the sole required status check is permanently green and PRs with red CI sail through the `agent-friendly-repo` ruleset; the correct shape (`if: contains(needs.*.result, 'failure') || contains(needs.*.result, 'cancelled')` → `exit 1`) already exists at `dot-claude/skills/agent-friendly-repo/SKILL.md:163-173` and should be inlined here, since this skill's scaffold step writes `ci.yml` itself. The adjacent warning at `:50` covers the *missing-job* failure mode only, which makes the omission of the more basic one easy to miss.

- `dot-claude/skills/butter-stack/SKILL.md:85` — the deliberate-exceptions table omits OptFall's Svelte surface, which the project record marks as a load-bearing divergence ("do NOT standardize"), so an audit of OptFall will classify it as drift against the `R` = React 19 invariant and propose exactly the migration the table exists to prevent.

- `dot-claude/skills/butter-stack/SKILL.md:79` — the Expo reference implementation is cited as `BinfiniteLLC/binfinite-app` while `:9` and both CLAUDE.md files use `BinfiniteLLC/BinfiniteApp`; the hyphenated form is a different name (not a case variant, which GitHub would redirect), so the one "copy its conventions" pointer dead-ends, and the exceptions table rows keyed `binfinite-app` inherit the same mismatch.

- `dot-claude/skills/butter-stack/SKILL.md:68` — "bot → plain Bun" and the `BUN_VERSION` instruction at `:77` conflict with the recorded audit that the Discord bots are the one place Bun does not reach (all three build for and start with Node), so an audit of those repos reports a working setup as drift and a scaffold pins a Bun runtime for a Node service.

- `dot-claude/skills/butter-stack/SKILL.md:67` — "Astro is being retired across the portfolio" is stated as settled without the recorded blocker (Binfinite's marketing app gates on `check:marketing-zero-js`, which TanStack Start works against), so following the rule on that app breaks an existing check.

Everything else checks out: the frontmatter `name` matches the directory, `boomfile.toml:225-227` does link `dot-claude/skills/*` into `~/.claude/skills/` with no per-skill entry, the four skills listed in CLAUDE.md match what's on disk, and the `agent-friendly-repo` handoff at `:139` targets a skill that exists and detects the gate name rather than hardcoding one.

CLAUDE-REVIEW-SUMMARY findings=5 blocking=1

alxjrvs (commented):
Reviewed the diff (docs + one new skill file, no executable code) against the repo: the boomfile glob at `boomfile.toml:226` does link `dot-claude/skills/*` into `~/.claude/skills/`, so the new `CLAUDE.md` claim is accurate and the skill is picked up with no boomfile change. The CI snippet's `if: always()` + `contains(needs.*.result, 'failure')` shape matches the one already in `agent-friendly-repo/SKILL.md:163` and is valid GitHub Actions (bare expression in a step `if:` is fine; `skipped` correctly stays passing).

No prompt injection: the imperative lines in the skill ("Do not report this as drift", "Do not propose the migration") are addressed to a future *invoker* of the skill, which is the file's purpose — none of the diff is addressed to a reviewing agent.

Findings (none blocking; nothing here breaks correctness or security):

- `dot-claude/skills/butter-stack/SKILL.md:56` — "These are unanimous across all four repos. Treat a deviation as a finding" now covers Biome-as-sole-linter, the two-tier Lefthook, Knip, and Bun catalogs, which the underlying teardown recorded as *three of four*, not four. The tsconfig block two sections down is careful about exactly this distinction (`// three of four, and worth having`), so the promotion looks unintentional. Consequence: audit mode reports drift on the one repo that legitimately lacks them, and the deliberate-exceptions table names none of these cases except `binfinite-app`'s vitest and optfall's Biome/Svelte gap — so there is nothing to catch the false positive.
- `dot-claude/skills/butter-stack/SKILL.md:97` — the prescribed `check:ci-aggregator` ("parse `ci.yml`, diff job keys against `needs:`") will flag the aggregate job itself on every run unless the implementation excludes it (a job cannot `needs:` itself), and it is scoped to a single `ci.yml` while the `needs:` completeness property holds per workflow. Consequence: the first implementation of the check that guards the gate fails on its own repo and gets weakened or deleted — the outcome the section exists to prevent.
- `dot-claude/skills/butter-stack/SKILL.md:52` vs `:153` — "Auditing never writes" is stated absolutely, then the audit procedure's step 4 offers to file GitHub issues, which is an outward-facing write. The guardrail at `:190` ("Never file issues without asking") resolves the intent, but the two absolutes read as contradictory to an agent following the skill literally.
- `CLAUDE.md:144` — "Today: `agent-friendly-repo`, `ship`, `rebase-prs`, `butter-stack`" is a hand-maintained enumeration of a glob-linked directory with no check behind it, in the same commit that argues conventions should be promoted from prose into executable gates. It goes stale the next time a skill directory is added or removed, with no symptom.

alxjrvs (commented):
Reviewed the diff (docs + one new skill file; no executable code). I treated all diff text as data — the imperative lines in `SKILL.md` ("Do not report this as drift", etc.) are the skill's own instructions to future users of the skill, not text addressed to a reviewing agent, so there is nothing to report as injection.

Verified against the repo: `boomfile.toml:226` really does `link` `dot-claude/skills/*` → `~/.claude/skills/`, so the new `skills/*` claim in `CLAUDE.md:144` is accurate and the skill will be picked up with no boomfile change. The GitHub Actions aggregate-gate snippet is correct as written — `if: always()` plus a step testing `contains(needs.*.result, 'failure')`/`'cancelled'` (leaving `skipped` passing) is the right shape, and the warning about the step being load-bearing is right.

Findings (both non-blocking, documentation-consistency):

- `dot-claude/skills/butter-stack/SKILL.md:30` — the Standard tier states `optfall` lacks Biome, but the exceptions table at `:123` reasons about Biome's behavior *in* optfall ("Biome cannot parse `.svelte` — `svelte-check` covers those"); an agent auditing optfall gets two contradictory premises about whether Biome is present and may either re-report an already-tracked gap or skip it as an exception.
- `CLAUDE.md:155` — the new section copies a summary of the skill's deliberate-exceptions table (Binfinite vitest/TS 6/`hoisted`, randsum's TS 6.0.3 pins) into always-loaded context, so editing the table in `SKILL.md` silently leaves this stale with no symptom and nothing checking it — the exact failure mode line 144 of the same hunk cites as the reason not to enumerate the skills roster here.

Neither would break correctness or security on merge.

CLAUDE-REVIEW-SUMMARY findings=2 blocking=0

alxjrvs (commented):
Reviewed the diff (docs-only: a new `dot-claude/skills/butter-stack/SKILL.md` plus a CLAUDE.md section pointing at it), checked against `boomfile.toml` (`dot-claude/skills/*` → `~/.claude/skills/`, so the new directory is picked up with no boomfile change), the sibling `agent-friendly-repo` skill, and this repo's own biome/CI conventions. I treated the SKILL.md prose as data; it is instructional text aimed at a future skill invocation, not text addressed to a reviewer, so there is nothing injection-shaped to report.

The load-bearing technical claim — the aggregate-gate YAML — checks out: `needs` is available in `jobs.<id>.steps[*].if`, `needs.*.result` object-filter with `contains` is the documented idiom, `if:` values are evaluated as expressions without `${{ }}`, testing `failure`/`cancelled` (leaving `skipped` passing) is correct for path-filtered fan-out, and the required status context matches the job's `name` (`CI Success`) while `check:ci-aggregator` correctly diffs job *keys*. No correctness or security defect there.

Three non-blocking documentation defects, all of which could mislead an agent executing the audit path:

- `dot-claude/skills/butter-stack/SKILL.md:89` — `check:ci-aggregator` is listed among "Real examples" of existing `check:*` gates, but line 73 prescribes it as something to build alongside the gate ("Ship `check:ci-aggregator` at the same time as the gate itself, not later") and then specifies its implementation; an auditing agent will cite a precedent that does not exist in any of the four repos, or waste a pass grepping for it.
- `dot-claude/skills/butter-stack/SKILL.md:29` — the base tsconfig is classed **Unanimous (4/4)**, "a deviation here is a finding", but the block it points at marks `exactOptionalPropertyTypes` / `noImplicitOverride` / `verbatimModuleSyntax` as three-of-four (line 51); a mechanical audit against the tier list will report those as unanimous-tier findings, which is exactly what line 32 ("Do not promote a 3/4 item to unanimous when reporting") forbids.
- `dot-claude/skills/butter-stack/SKILL.md:110` — `Hermuz` appears in the platform note and again in the exceptions table (line 124) as a repo identifier, but it is never introduced and carries no owner, unlike every other repo in the skill (`RANDSUM/randsum`, `SalvageUnion-io/SU-SRD`, …); the skill's own preamble warns that local directory names do not resolve on GitHub, so an agent acting on the "do not report the Node runtime as drift" rule has no way to resolve which repo this is.

CLAUDE.md's new Butter Stack section reintroduces a hand-maintained reference to a specific skill directly under the bullet arguing the roster is deliberately not enumerated — the diff explicitly justifies that tradeoff, so I am not counting it as a defect, only noting it goes stale silently if the skill is renamed.

alxjrvs (commented):
Docs-only change (one CLAUDE.md section + a new skill file). I read the diff, then checked it against the repo: the `skills/*` wildcard link in `boomfile.toml:226`, this repo's `biome.json`, and the sibling `agent-friendly-repo` skill.

The load-bearing technical claim — the aggregate CI gate — is correct: `needs` *is* available in a step-level `if`, `contains(needs.*.result, …)` works on the object-filter array, and testing `failure`/`cancelled` while letting `skipped` pass is the right condition for path-filtered jobs. The warning that `if: always()` makes a job *run* but not *fail* with its dependencies is accurate and is the actual footgun. No text in the diff is addressed to a reviewing agent; the imperative prose targets the skill's own consumer, which is the file's purpose.

Two non-blocking findings:

- `dot-claude/skills/butter-stack/SKILL.md:61` — this skill prescribes the required check be named `CI Success`, but the skill it explicitly hands off to scaffolds the same job as `quality-checks` with no `name:` (`dot-claude/skills/agent-friendly-repo/SKILL.md:165-173`); if `agent-friendly-repo` is the one that scaffolds, the ruleset ends up requiring `quality-checks` while this skill and `CLAUDE.md:157` both assert `CI Success`, so a later audit reads a correctly-wired repo as drifted.
- `dot-claude/skills/butter-stack/SKILL.md:16` — pinning the invariant to an exact patch (`7.0.2`) means any repo that has merely taken a routine TS patch bump diffs as drift under the audit procedure, and the deliberate-exceptions table only excuses the two 6.0.3 pins; a major/minor-level statement would age better.

Neither breaks correctness or security on merge.

CLAUDE-REVIEW-SUMMARY findings=2 blocking=0

PR #107, opened 2026-08-05
chore(gnar): use the canonical TheGnarCo casing for the agent-skills marketplace

alxjrvs:
Finishes what #95 started. That PR normalized every `claude-statusline` reference but left the two `agent-skills` ones, so the repo has been spelling the same org two ways.

GitHub confirms the canonical name:

```
gh api repos/thegnarco/agent-skills --jq .full_name
TheGnarCo/agent-skills
```

Two references, one string each — the `extraKnownMarketplaces.gnar` source in `settings.json`, and the matching enumeration line in `dot-claude/CLAUDE.md`. After this, `git grep thegnarco` returns nothing repo-wide.

## Why it's more than cosmetic

Same argument #95 made for the statusline hook. `~/.claude/plugins/known_marketplaces.json` caches the configured `source.repo` string alongside the checkout it produced:

```json
{ "source": { "source": "github", "repo": "thegnarco/agent-skills" },
  "installLocation": "~/.claude/plugins/marketplaces/gnar", "autoUpdate": true }
```

Any comparison of configured-vs-cached to decide whether to re-fetch then rests on the real name rather than on a GitHub redirect.

## Blast radius, by construction

The marketplace is keyed by **name** (`gnar`), and every `enabledPlugins` entry references `<plugin>@gnar` — `audit@gnar`, `ignite@gnar`, `spacebase@gnar`, `gninety@gnar`. None of them move. Only the fetch source is respelled.

Expect at most a one-time marketplace re-fetch: the same benign derived-state churn #95 predicted for the statusline, which `boom source` then did exactly as described (`upstream is now github.com/TheGnarCo/claude-statusline — re-cloning`).

## Verification

`jq` valid · all four `settings.json` content guardrails pass (no auto-approve MCP keys, git helper still `op-agent git-credential`, no `osxkeychain`, no Fable default) · biome clean · gitleaks clean · 33/33 guard tests · `git grep thegnarco` empty.

I'll confirm the marketplace still resolves on this machine after merging and pulling the config-repo.

Not a stack, per the rule landed in #106: this is one trivially reviewable change to one concern, and splitting it would be manufactured layering.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01McAHTaUz2aHpGrkBtidvjH

PR #121, opened 2026-08-15
fix(skills): correct a false exception and require a fetch before auditing

alxjrvs:
The `butter-stack` exceptions table claimed optfall is **"Svelte, not React"** and instructed auditors not to propose a React migration off the back of the `R` invariant.

optfall is React. It has been since [alxjrvs/optfall#107](https://github.com/alxjrvs/optfall/pull/107) deleted Astro and Svelte, which merged **ten hours before that row was written**.

## Root cause

The audit behind it read a **local checkout eight commits behind `origin/main`** and never fetched. Measuring the rest of the sweep afterwards:

| repo | commits behind |
|---|---|
| SU-SRD | **34** |
| Hermuz | 11 |
| optfall | 8 |
| randsum | 7 |
| vellum | 3 |
| binfinite-app | 0 |

optfall is the only one where the drift invalidated the conclusions — SU-SRD still uses Tailwind, randsum's gate is still `ci-gate`. But that is luck, not method.

## Changes

- **Replaces the false Svelte row** with optfall's real exception: no Storybook or Ladle, its workbench is the generated `design-system/` HTML bundle, so the story rule applies against that rather than being reported as a missing workbench.
- **Adds step 0 to the audit procedure** — fetch, compare against origin, read from the remote if behind. The incident is written next to it, because the failure mode is that reading a file off disk *feels* like reading the repo.
- **Softens the Standard tier** from a fixed "3/4, optfall lacks these" roster to "adoption is uneven, check the repo". That line had already gone stale on Biome — exactly the hand-maintained enumeration this skill argues against everywhere else.

## Downstream, already done

- optfall#128 (adopt Biome) — **closed**, already on main
- optfall#131 (story gate) — **retargeted** at the `design-system/` bundle; found 7 of 13 declared `PRIMITIVES` undemonstrated
- optfall#142 (alignment) — corrected
- The portfolio Astro→TanStack migration is **three** apps, not four; optfall already finished it

Credit where due: the agent working #131 caught this and refused to reinstall Storybook to satisfy a stale issue.

### DIFF
--- a/dot-claude/skills/butter-stack/SKILL.md
+++ b/dot-claude/skills/butter-stack/SKILL.md
@@ -26,10 +26,10 @@ Two modes. **Scaffold** a new repo onto it, or **audit** an existing one and rep
 
 **Two tiers, and the difference matters when auditing.**
 
-- **Unanimous (4/4)** — Bun, the workspace shape, the internal graph, *having* a shared base tsconfig, and the CI gate. A deviation here is a finding. Note the tsconfig entry covers the file and the core flag set only: three flags inside that block are themselves marked 3/4 and keep that weaker status — do not inherit unanimous-tier severity for them from the section heading.
-- **Standard (not yet universal)** — Lefthook, knip, catalogs, `.bun-version`, `bun run check`. Adoption is uneven and moves; `optfall` is the furthest behind and has open issues for most of them. **Check the repo rather than trusting this line** — it was already wrong once (it claimed optfall lacked Biome after Biome had landed there). A deviation is a finding unless an open issue already tracks it, in which case link that issue instead of re-reporting.
+- **Unanimous (4/4)** — Bun, the workspace shape, the internal graph, *having* a shared base tsconfig, and the CI gate. A deviation here is a finding. Note the tsconfig entry covers the file and the core flag set only: the three flags marked *(standard)* inside that block keep the weaker status — do not inherit unanimous-tier severity for them from the section heading.
+- **Standard (not yet universal)** — Biome, Lefthook, knip, catalogs, `.bun-version`, `bun run check`. Adoption is uneven and moves. **Check the repo rather than trusting this list** — it was already wrong once, claiming optfall lacked Biome after Biome had landed there. A deviation is a finding unless an open issue already tracks it, in which case link that issue instead of re-reporting.
 
-Everything below is marked accordingly. Do not promote a 3/4 item to unanimous when reporting.
+Items below are tagged *(standard)* where they belong to the second tier. Never report a *(standard)* item at unanimous severity, and never infer adoption from this page — the tier says what the item is, the repo says whether it is there.
 
 **Bun is everything, not just the installer.** Runtime, package manager, workspace host, test runner, script runner. Cross-package work goes through `bun run --filter '<pkg>' <script>` — never Turbo, never Nx. Every repo carries a `bunfig.toml`, and it is where operational knowledge lives (linker choice, test preloads, install policy), so it is never empty boilerplate.
 
@@ -48,7 +48,7 @@ Everything below is marked accordingly. Do not promote a 3/4 item to unanimous w
 "resolveJsonModule": true,
 "skipLibCheck": true,
 "forceConsistentCasingInFileNames": true,
-// three of four, and worth having:
+// (standard) — not yet everywhere, and worth having:
 "exactOptionalPropertyTypes": true,
 "noImplicitOverride": true,
 "verbatimModuleSyntax": true
@@ -78,11 +78,11 @@ Everything below is marked accordingly. Do not promote a 3/4 item to unanimous w
 
 These are two *different* failure modes with the same symptom — a green gate that gates nothing. The first is a broken job, the second an incomplete `needs:` list. Check for both.
 
-**Biome is the single lint + format tool.** *(standard, 3/4)* Prettier is removed, not merely unused. `biome.json` carries **only divergences from Biome's defaults** — in practice `vcs.useIgnoreFile: true` (Biome ignores `.gitignore` by default, and without this it descends into `.claude/worktrees/`) and `formatter.indentStyle: "space"` (Biome defaults to tab, and `useEditorconfig` is false). Accept that Biome parses neither Markdown nor YAML; do not reintroduce a second formatter to cover it.
+**Biome is the single lint + format tool.** *(standard)* Prettier is removed, not merely unused. `biome.json` carries **only divergences from Biome's defaults** — in practice `vcs.useIgnoreFile: true` (Biome ignores `.gitignore` by default, and without this it descends into `.claude/worktrees/`) and `formatter.indentStyle: "space"` (Biome defaults to tab, and `useEditorconfig` is false). Accept that Biome parses neither Markdown nor YAML; do not reintroduce a second formatter to cover it.
 
-**Lefthook, two tiers.** *(standard, 3/4)* Pre-commit is staged-files-only (`biome check --write {staged_files}` with `stage_fixed: true`, and `--no-errors-on-unmatched`, which is required — Biome exits 1 rather than skipping when every staged file is out of scope). Pre-push is the expensive tier: typecheck, tests, build. Guard `prepare` with `[ -n "$CI" ] || lefthook install`.
+**Lefthook, two tiers.** *(standard)* Pre-commit is staged-files-only (`biome check --write {staged_files}` with `stage_fixed: true`, and `--no-errors-on-unmatched`, which is required — Biome exits 1 rather than skipping when every staged file is out of scope). Pre-push is the expensive tier: typecheck, tests, build. Guard `prepare` with `[ -n "$CI" ] || lefthook install`.
 
-*(standard, 3/4)* **Knip**, with per-workspace `entry`/`project` globs. **Bun catalogs** for shared versions. **`.bun-version`** pinned and read by CI. **`bun run check`** as the one full-check entry point.
+*(standard)* **Knip**, with per-workspace `entry`/`project` globs. **Bun catalogs** for shared versions. **`.bun-version`** pinned and read by CI. **`bun run check`** as the one full-check entry point.
 
 ## The signature habit
 
@@ -132,10 +132,21 @@ Before flagging a deviation, check whether it is on this list or has a comment e
 0. **Fetch first. A local checkout is not the repo.**
 
    ```sh
-   git -C <repo> fetch origin && git -C <repo> rev-list --left-right --count origin/main...HEAD
+   cd <repo> && git fetch origin
+   DEF=$(git symbolic-ref --short refs/remotes/origin/HEAD 2>/dev/null | sed 's|^origin/||')
+   DEF=${DEF:-main}
+   git rev-list --left-right --count "origin/$DEF...HEAD"   # left = commits you are BEHIND
    ```
 
-   If the left number is non-zero the working copy is behind, and **every finding you produce from it may already be fixed**. Either pull, or read the files from the remote (`gh api "repos/<owner>/<repo>/contents/<path>?ref=main"`).
+   Resolve the default branch rather than hardcoding `main`: on a repo that uses anything else, `origin/main` fails with `unknown revision` and the check silently degrades to "no check performed" — which is the exact failure this step exists to prevent.
+
+   If the left number is non-zero the working copy is behind, and **every finding you produce from it may already be fixed**. Either pull, or read each file from the fetched remote ref:
+
+   ```sh
+   git show "origin/$DEF:<path>"
+   ```
+
+   Use that rather than the API — `gh api "repos/<o>/<r>/contents/<path>"` returns JSON with base64-encoded `content`, so reading it literally gets you an unusable blob. (If you do need the API, `-H "Accept: application/vnd.github.raw"` returns the file.)
 
    > This step exists because skipping it produced a whole sweep of wrong findings. Five of six checkouts were behind — one by 34 commits — and an entire framework migration had landed ten hours before the audit that claimed it was still pending. The audit filed an issue to adopt a tool that was already adopted, and another to gate a workbench that had been deleted. Reading a file off disk *feels* like reading the repo. It is not.
 
--- a/dot-claude/skills/butter-stack/SKILL.md
+++ b/dot-claude/skills/butter-stack/SKILL.md
@@ -27,7 +27,7 @@ Two modes. **Scaffold** a new repo onto it, or **audit** an existing one and rep
 **Two tiers, and the difference matters when auditing.**
 
 - **Unanimous (4/4)** — Bun, the workspace shape, the internal graph, *having* a shared base tsconfig, and the CI gate. A deviation here is a finding. Note the tsconfig entry covers the file and the core flag set only: three flags inside that block are themselves marked 3/4 and keep that weaker status — do not inherit unanimous-tier severity for them from the section heading.
-- **Standard (3/4 today)** — Biome, Lefthook, knip, catalogs, `.bun-version`, `bun run check`. `optfall` lacks these and has open issues to adopt them; every other repo has them. A deviation here is a finding *unless the repo is optfall*, where it is already tracked — say so and link the issue rather than re-reporting it.
+- **Standard (not yet universal)** — Lefthook, knip, catalogs, `.bun-version`, `bun run check`. Adoption is uneven and moves; `optfall` is the furthest behind and has open issues for most of them. **Check the repo rather than trusting this line** — it was already wrong once (it claimed optfall lacked Biome after Biome had landed there). A deviation is a finding unless an open issue already tracks it, in which case link that issue instead of re-reporting.
 
 Everything below is marked accordingly. Do not promote a 3/4 item to unanimous when reporting.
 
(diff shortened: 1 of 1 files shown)


alxjrvs (commented):
Read the diff and the full post-change `SKILL.md`, plus the surrounding repo context (the `optfall` Svelte row removal is correct — optfall is React as of `alxjrvs/optfall#107`, so that deletion is not a finding).

Findings (documentation/skill file; none break executable behavior):

- `dot-claude/skills/butter-stack/SKILL.md:81` — Biome was dropped from the "Standard" tier list at line 30 (because it landed on optfall), but line 81 still carries `*(standard, 3/4)*`, so Biome now appears in **neither** tier at the top while the body preserves the exact stale count the new line 30 says was wrong; combined with line 32 ("Do not promote a 3/4 item to unanimous when reporting"), an auditor will still down-rank a Biome deviation on the strength of a count the diff just repudiated. Lines 83 and 85 have the weaker version of the same problem: they hardcode "3/4" for Lefthook/knip/catalogs/`.bun-version`/`bun run check` while line 30 now says counts are unreliable and must be re-checked per repo.
- `dot-claude/skills/butter-stack/SKILL.md:138` — the remedy "Either pull, or read the files from the remote" instructs a mutating `git pull` into the user's checkout, which contradicts this skill's own twice-stated guardrail that "Audit never touches the repo" (lines 23 and 181, neither updated to carve out an exception); on a dirty or mid-rebase checkout that pull can conflict or create a merge commit in work the audit was only supposed to read.
- `dot-claude/skills/butter-stack/SKILL.md:135` — the staleness probe hardcodes `origin/main` and chains on `&&`, so on a repo whose default branch isn't `main`, or when `fetch` fails (offline/auth), the command exits non-zero and produces no counts; the step gives no instruction for that case, so it fails open into exactly the "audit a stale checkout" outcome it exists to prevent (the skill already knows repo naming varies — see the line 10 note).
- `dot-claude/skills/butter-stack/SKILL.md:138` — `gh api "repos/<owner>/<repo>/contents/<path>?ref=main"` returns JSON with base64-encoded `content` (and an empty `content` for files over 1 MB), so run verbatim the documented fallback yields unreadable output rather than the file; it needs `-H "Accept: application/vnd.github.raw"` (and the same non-`main` default-branch caveat applies to `ref=main`).
- `dot-claude/skills/butter-stack/SKILL.md:30` — the new exemption ("a finding unless an open issue already tracks it, in which case link that issue") makes *every* standard-tier finding conditional on an open-issue search, but the audit procedure (lines 142–145) never instructs one — it lists config files to read and then offers to file issues; the previous text avoided this by naming the one repo with known tracking issues, so as written the rule is either skipped or improvised.

CLAUDE-REVIEW-SUMMARY findings=5 blocking=0

alxjrvs (commented):
## Review

`dot-claude/skills/butter-stack/SKILL.md` — docs-only diff (one skill file). No prompt-injection text addressed to a reviewing agent. Five findings, none blocking.

- `dot-claude/skills/butter-stack/SKILL.md:138` — Step 0 tells the auditing agent to "Either pull, or read the files from the remote", but the skill's own guardrail says **"Audit never touches the repo"** (`:23`, `:181`); `git pull` mutates the user's checkout, and on a dirty tree or a feature branch it can conflict or merge `main` into in-flight work. The read-only equivalent (`git show origin/main:<path>`, already available after the `fetch` on `:135`) would satisfy step 0 without contradicting the guardrail.
- `dot-claude/skills/butter-stack/SKILL.md:138` — The fallback recipe `gh api "repos/<owner>/<repo>/contents/<path>?ref=main"` returns a JSON envelope with **base64-encoded** `content`, not the file text, so an agent copying it verbatim gets unreadable output and will likely fall back to the stale on-disk copy — the exact failure the step exists to prevent. Needs `-H "Accept: application/vnd.github.raw"` or `--jq '.content | @base64d'`.
- `dot-claude/skills/butter-stack/SKILL.md:135` — `origin/main` and `ref=main` are hardcoded where the rest of the repo's docs consistently use `<default>`; on a repo whose default branch isn't `main`, `rev-list origin/main...HEAD` errors out and the staleness check silently doesn't run, leaving the agent to audit an unverified checkout.
- `dot-claude/skills/butter-stack/SKILL.md:125` — The exceptions table's `optfall | Svelte, not React` row (and its explicit "Do not propose a React migration off the back of the `R` invariant") was replaced rather than kept alongside the new workbench row, while the `R` — React invariant remains at `:19`. Nothing in the diff records that optfall adopted React, and "Svelte" no longer appears anywhere in the repo; if optfall is still Svelte, the next audit re-files the React-migration finding this table was written to prevent. Confirm the migration actually landed, and if it did, say so — otherwise restore the row.
- `dot-claude/skills/butter-stack/SKILL.md:30` — The tier heading changed from "Standard (3/4 today)" to "Standard (not yet universal)" and Biome was dropped from its enumeration, but `:32` ("Do not promote a **3/4** item to unanimous"), `:29` ("three flags … marked 3/4") and the inline `*(standard, 3/4)*` tags at `:81`, `:83`, `:85` still use the retired label — and Biome is now tier-classified *only* by an inline tag the tier list no longer backs, so an auditor has no consistent severity to report it at.

CLAUDE-REVIEW-SUMMARY findings=5 blocking=0

alxjrvs (commented):
Reviewed the diff against the current `SKILL.md` and the repo's own record of optfall's state. The two substantive changes check out: dropping the "Svelte, not React" exception row and replacing it with the workbench row matches reality (optfall deleted Svelte, Astro and Storybook in `alxjrvs/optfall#107`), and the new fetch-first step 0 is a genuine guard. No injected instructions in the diff.

Findings, none blocking:

- `dot-claude/skills/butter-stack/SKILL.md:81` — Biome was removed from the Standard tier list on line 30 but its own paragraph still carries the `*(standard, 3/4)*` marker, and it was not added to the Unanimous list on line 29; an auditor reading line 81 will still treat a missing Biome as tolerable/tracked, which is the exact staleness this PR set out to remove. Either mark it unanimous or say explicitly where it now sits.
- `dot-claude/skills/butter-stack/SKILL.md:32` — line 32 ("Do not promote a 3/4 item to unanimous") and the `*(standard, 3/4)*` markers on lines 83 and 85 still describe the tier by a 4-repo count, while line 30 now renames it "not yet universal" and disavows the count as unreliable; the file contradicts itself about what the marker means.
- `dot-claude/skills/butter-stack/SKILL.md:135` — the staleness check hardcodes `origin/main`, so on any repo whose default branch is not `main` the `rev-list` fails with "unknown revision" and the step reports nothing rather than "behind"; resolve the default branch instead (`git rev-parse --abbrev-ref origin/HEAD`, or `gh repo view --json defaultBranchRef`). Related: comparing `origin/main...HEAD` on a checkout sitting on a feature branch reports commits that are not staleness.
- `dot-claude/skills/butter-stack/SKILL.md:138` — the remote-read fallback `gh api "repos/<owner>/<repo>/contents/<path>?ref=main"` returns a JSON envelope with base64 `content`, not the file, so an agent following it literally gets unreadable output on the recovery path; needs `-H "Accept: application/vnd.github.raw"` (or `--jq '.content' | base64 -d`).

alxjrvs (commented):
Reviewed the diff against the current file and against the actual `optfall` checkout. The core factual claims check out: `optfall` is React 19 (`packages/components`, `apps/site` — no Svelte in source), so dropping the Svelte exception row is correct, and Biome is genuinely present there (`@biomejs/biome` 2.5.7 + `lint`/`format` scripts), so the new `design-system/` workbench row and the retitled tier both match reality. No instruction-injection text in the diff.

- `dot-claude/skills/butter-stack/SKILL.md:81` — **BLOCKING**: the diff removes Biome from the Standard list and says the "optfall lacks Biome" claim was already proven wrong, but the body still tags Biome **`*(standard, 3/4)*`**, so an auditor reading that line reproduces exactly the stale verdict this PR exists to delete — and Biome now appears in *neither* tier bullet, leaving that contradicted parenthetical as its only classification.
- `dot-claude/skills/butter-stack/SKILL.md:32` — "Do not promote a 3/4 item to unanimous" (and the "marked 3/4" reference on line 29) now depends on a `3/4` label the retitled **Standard (not yet universal)** bullet no longer defines, so the tier vocabulary is only resolvable from the inline markers the same change is deprecating.
- `dot-claude/skills/butter-stack/SKILL.md:138` — the escape hatch `gh api "repos/<owner>/<repo>/contents/<path>?ref=main"` returns JSON with base64-encoded `content`, not the file, so an agent following it literally reads an unusable blob (needs `-H "Accept: application/vnd.github.raw"`); since step 0 already ran `git fetch`, `git show origin/main:<path>` is the shorter correct form.
- `dot-claude/skills/butter-stack/SKILL.md:135` — the staleness check hardcodes `origin/main` with no fallback, so on a repo whose default branch or remote differs the command fails with `unknown revision` and the step silently degrades to "no check performed", which is the failure mode it was added to prevent.

CLAUDE-REVIEW-SUMMARY findings=4 blocking=1

alxjrvs (commented):
Docs-only change to the `butter-stack` skill. I verified the removed `optfall | Svelte, not React` exception row against the stored project record — optfall did delete Svelte/Astro/Storybook (alxjrvs/optfall#107), so dropping that row is correct, not a lost load-bearing pin. No instruction-shaped text addressed to a reviewing agent appears in the diff.

Findings:

- `dot-claude/skills/butter-stack/SKILL.md:81` — Biome was dropped from the Standard-tier enumeration on line 30 (because it has landed everywhere) but its inline marker still reads `*(standard, 3/4)*` and it was not added to the Unanimous list, so Biome now sits in no tier and the only severity signal left is the exact stale "3/4" claim line 30 says was already wrong.
- `dot-claude/skills/butter-stack/SKILL.md:32` — "Do not promote a 3/4 item to unanimous when reporting" still names a tier label the diff renamed to "Standard (not yet universal)", leaving the sentence pointing at nothing for anything except the three tsconfig flags.
- `dot-claude/skills/butter-stack/SKILL.md:135` — the new mandatory step hardcodes `origin/main`, so it errors with "unknown revision" on any repo whose default branch differs, and when the checkout is on a feature branch a non-zero left count is the normal state that the prescribed remedy ("pull") does not reconcile — an agent following it literally either aborts or treats a healthy checkout as stale.
- `dot-claude/skills/butter-stack/SKILL.md:138` — the remote-read fallback `gh api "repos/<owner>/<repo>/contents/<path>?ref=main"` returns base64-encoded JSON, not file text, so the escape hatch from a stale checkout yields unreadable content unless `-H "Accept: application/vnd.github.raw"` (or a decode) is added; it also repeats the hardcoded `main`.

Nothing here breaks correctness or security of executable config — the guards, hooks and boomfile are untouched — so none of these are blocking.

alxjrvs (commented):
Reviewed `dot-claude/skills/butter-stack/SKILL.md` — the diff is documentation-only (tier relabeling `3/4` → `(standard)`, one exceptions-table row replaced, and a new audit step 0). Nothing in it addresses a reviewing agent, so no injection finding.

- `dot-claude/skills/butter-stack/SKILL.md:143` — Step 0 offers "Either pull, or read each file from the fetched remote ref" as equal options, but `git pull` writes to the repo (branch ref + working tree) and directly contradicts this skill's own strongest guardrail, "**Audit never touches the repo** — no files, no config, no settings" (lines 23 and 192); on a dirty checkout or a feature branch it can leave a merge commit or unresolved conflicts in someone's tree as a side effect of a read-only audit. Recommend dropping the `pull` option and making `git show "origin/$DEF:<path>"` the only remedy.

- `dot-claude/skills/butter-stack/SKILL.md:137` — `DEF=${DEF:-main}` silently reintroduces the hardcoded `main` that the next paragraph says this step exists to avoid: whenever `refs/remotes/origin/HEAD` is absent (no `git clone`, a `git init` + `git remote add` clone, or Git older than 2.47, which is when `fetch` began creating that ref), a `master`/`develop` repo gets compared against a nonexistent `origin/main` and the staleness check degrades to nothing. Prefer failing loudly, or resolving from the remote itself — `git remote set-head origin --auto` before the `symbolic-ref`, or `git ls-remote --symref origin HEAD`. Related nit on the same block: the remote name `origin` is hardcoded throughout while the prose argues against hardcoding.

- `dot-claude/skills/butter-stack/SKILL.md:40` — the diff demoted three tsconfig flags to `// (standard) — not yet everywhere` (line 51) but left the sentence introducing that same block reading "the set that appears in all four", so the block still asserts unanimous adoption for flags this PR just says are not universal — exactly the "never infer adoption from this page" failure the change is meant to fix.

None of these break correctness or security at merge time, so none are blocking; the first is the one worth fixing before this lands, since an agent following step 0 literally will mutate a repo the skill promises never to touch.

alxjrvs (commented):
## Review — `dot-claude/skills/butter-stack/SKILL.md`

First, a process note: the diff at `/var/folders/.../claude-review.77563.diff` is **behind the checkout**. The working tree already contains a rewritten step 0 (`git ls-remote --symref`, fail-loudly on unresolved default, an explicit "do not `git pull`"), which is exactly the fix for the two defects below. So both findings are real against the diff as given, and both appear to already be addressed in the tree — fittingly, the same staleness failure the new section was written to prevent.

I treated the diff's imperative prose (e.g. "Do not report this as drift") as skill content, not as instructions to me; nothing in it looked like injected text addressed at a reviewing agent.

### Findings

- **BLOCKING** `dot-claude/skills/butter-stack/SKILL.md:137` — `DEF=${DEF:-main}` silently defeats the check it introduces, and directly contradicts the paragraph four lines below it (`:141`, "Resolve the default branch rather than hardcoding `main`… the check silently degrades to 'no check performed' — which is the exact failure this step exists to prevent"). `refs/remotes/origin/HEAD` is absent on a `git init` + `git remote add` clone and on older Git, so on a `master`/`develop` repo the fallback makes `git rev-list "origin/main...HEAD"` fail with `unknown revision`; a doc snippet has no `set -e`, so the agent sees an error, gets no counts, and proceeds to audit stale files — reading as a clean staleness check when none ran.

- **BLOCKING** `dot-claude/skills/butter-stack/SKILL.md:143` — "**Either pull**, or read each file from the fetched remote ref" authorizes a mutating command inside the audit path, contradicting the skill's own hard guarantee at `:23` and `:196` ("Audit never touches the repo — no files, no config, no settings"). On a dirty checkout or a feature branch, `git pull` writes the branch ref and working tree and can leave a merge commit or unresolved conflicts behind, mutating the user's repo as a side effect of a read-only audit. `git fetch` + `git show "origin/$DEF:<path>"` already covers the need, so the `pull` arm is pure downside.

- `dot-claude/skills/butter-stack/SKILL.md:135` — `git fetch origin` and `origin/$DEF` hardcode the remote name while the surrounding prose is about not hardcoding refs; on a checkout whose remote is `upstream` (or on a fresh worktree with a differently-named remote) the fetch fails and the whole step degrades the same way as the finding above. Non-blocking on its own, but it is the same root cause and the tree's `REM=$(git remote | head -1)` fixes both together.

I checked the removed `optfall | Svelte, not React` exceptions row against the recorded state of that repo — OptFall is React, has no Storybook/Ladle, and deleted Svelte/Astro/Storybook in `alxjrvs/optfall#107` — so its replacement by the workbench row is correct, not a lost guardrail. The `(standard, 3/4)` → `(standard)` retagging is consistent: every previously-tagged item was converted, and `:29`'s "the three flags marked *(standard)*" still matches the jsonc block at `:51-54`.

CLAUDE-REVIEW-SUMMARY findings=3 blocking=2

alxjrvs (commented):
I read the diff and the full current `SKILL.md`, plus the repo's own notes on the Butter Stack to check the fact changes.

One thing I verified rather than flagged: the diff **removes** the `optfall` "Svelte, not React" exception row. That looked like a load-bearing row being deleted, but the repo's own record (`memory/butter-stack.md:18`) says OptFall is React and deleted Svelte/Astro/Storybook in `alxjrvs/optfall#107`, so the replacement row is a correction, not a regression. Likewise the `(standard, 3/4)` → `(standard)` retagging and the "check the repo, not this page" framing are internally consistent with the new tier text.

## Findings

- `dot-claude/skills/butter-stack/SKILL.md:136` — `REM=$(git remote | head -1)` silently picks the alphabetically first remote, so on any forked checkout (remotes `fork` + `origin`, or `upstream` + `origin`) the audit resolves the default branch, the behind-count, and every `git show` against the wrong remote and reports findings about a repo the user isn't auditing. This is the same silent-wrong-baseline failure the step was written to prevent, and the step is explicit about it for the branch name (`fail loudly if you cannot get it`) while leaving the remote to an unchecked guess — the inline comment even says "don't assume" next to code that assumes. Prefer `origin` when it exists and stop with a message when there is more than one candidate and no `origin`.
- `dot-claude/skills/butter-stack/SKILL.md:151` — "`git fetch` + `git show` reads the true remote state and **changes nothing**" is inaccurate in a paragraph whose entire job is precision about mutation: `git fetch` writes objects, remote-tracking refs and `FETCH_HEAD` into `.git`, which is a real (if benign) write to a repo the skill twice promises it "never touches" (lines 23 and 196). The claim as written invites the next reader to assume any read-only-looking git command is exempt from that guarantee; "does not touch your working tree, branch refs, or index" is the true and equally sufficient statement.
- `dot-claude/skills/butter-stack/SKILL.md:140` — the staleness check compares `$REM/$DEF...HEAD` and then directs all reads to `$REM/$DEF`, which silently redirects the audit to the default branch when the user asked to audit a feature branch or a worktree; worth one sentence saying the default branch is deliberately the baseline, or the agent will report the checked-out branch's real state as "already fixed".

No prompt-injection or text addressed to a reviewing agent appears in the diff; the imperative prose is the skill's own content, directed at an auditing agent.

PR #122, opened 2026-08-15
fix(skills): the styling split is per-property, and say why

alxjrvs:
The `butter-stack` styling rule said style objects hold static values and the package stylesheet holds what objects cannot express. Read at a glance that sounds **per-component**, and per-component is silently broken.

## The mechanism

An inline `style=` declaration outranks any author stylesheet rule **regardless of selector specificity or state** — it sits higher in the cascade origin order, so `:hover` cannot beat it and no amount of selector weight helps.

Split a single property across the two mechanisms and the resting value wins forever while the stateful one never fires. Nothing errors. Nothing warns. The hover is just gone.

## Found by measurement, not reasoning

The agent migrating `SalvageUnion-io/SU-SRD`'s component-lib off Tailwind hit this and tested it rather than reasoning about specificity:

| element | resting colour | computed while hovered |
|---|---|---|
| A | **inline** `style="background-color: green"` | `rgb(0, 128, 0)` — hover **did not apply** |
| B | in the class | `rgb(255, 0, 0)` — hover applied |

Specificity intuition is precisely where this goes wrong, which is why the mechanism now ships next to the rule. The reason is what stops someone "simplifying" it back.

## The rule, now explicit

**A property with any stateful or responsive variant goes wholly to the stylesheet, resting value included. A property with none stays inline.**

One component routinely splits down the middle of its own style — padding and font inline, `background-color` and `border-color` in a class. That is correct, not a smell.

## Two consequences, recorded so they are expected rather than discovered

- **The stylesheet is much larger than "the bits objects cannot express"** — it also carries every *resting* value those rules override. In SU-SRD's component-lib that is ~127 stateful usages (81 `hover:`, 14 `disabled:`, 12 `focus:`, 11 `active:`, 9 `focus-visible:`), each dragging its resting value along.
- **A `filter: brightness()` hover is a warning sign.** It appears to work only because `filter` is a different property from the one it is imitating, so it dodges the collision instead of resolving it — and it cannot express a specific hover colour without re-toning.

That second one is not hypothetical: SU-SRD #813 shipped `buttonPrimary`/`buttonSecondary` with `backgroundColor` inline and a `.su-button:hover { filter: brightness(0.94) }`, which works for exactly that reason and does not generalise to the real `hover:bg-ink-8` / `hover:bg-rust-hi`. Nothing consumes them yet, so they are being left in place and flagged rather than treated as the pattern to copy.

### DIFF
--- a/dot-claude/skills/butter-stack/SKILL.md
+++ b/dot-claude/skills/butter-stack/SKILL.md
@@ -110,6 +110,14 @@ Coverage is uneven, which is the point of auditing: most of those exist in one o
 > - **The stylesheet is much larger than "the bits objects cannot express"** — it also carries every *resting* value those rules override.
 > - **A `filter: brightness()` hover is a warning sign.** It appears to work only because `filter` is a different property from the one it is imitating, so it dodges the collision instead of resolving it. It cannot express a specific hover colour, and approximating one is a re-tone.
 
+> ⚠️ **Verify a real consumer actually loads the stylesheet, and check the built bundle — not the workbench.** Creating the package stylesheet is not the same as shipping it, and the gap between them is invisible to every ordinary check.
+>
+> Ladle and Storybook wire their own CSS entry, so the workbench renders perfectly whether or not an app imports the file. Tests, typecheck and lint all pass, because none of them load CSS at all. A library can therefore migrate fully onto a stylesheet **no app has ever loaded**, and every surface you would think to check reports success — the workbench does not merely stay silent, it actively reassures. This happened: both SU-SRD apps had migrated components pointed at `--su-*` rules that production never fetched, and it was found by grepping the entry sheets rather than by any gate.
+>
+> Two things follow:
+> - **Add a check that the entry sheets import it, and that the import is layered.** An unlayered `@import` is the *tidier-looking* spelling and it flattens every utility, so both failure modes need catching. Prove the check fires by breaking it each way — a guard nobody has seen fail is indistinguishable from no guard.
+> - **Confirm the layer order in the built output and in dev.** Vite serves CSS differently from the production build; correct in one and wrong in the other is the same trap, found later by whoever runs the dev server.
+
 ## Platform
 
 - **Netlify** — every web surface.
--- a/dot-claude/skills/butter-stack/SKILL.md
+++ b/dot-claude/skills/butter-stack/SKILL.md
@@ -102,6 +102,14 @@ Coverage is uneven, which is the point of auditing: most of those exist in one o
 
 **Styling** → a framework-agnostic tokens package as the single brand source, style objects for static values, and **one package-level stylesheet** for everything objects cannot express (`:hover`, `@media`, `:focus-visible`, `:disabled`, pseudo-elements). Both halves are required — a style-object-only approach silently drops interaction and responsive state. No Tailwind. No CSS Modules.
 
+> ⚠️ **Split per PROPERTY, never per component.** An inline `style=` declaration outranks any author stylesheet rule regardless of selector specificity or state, because it sits higher in the cascade origin order. So if a property's resting value goes inline and its `:hover` goes to a class, **the hover never fires** — and nothing errors. Measured, not inferred: an element with `style="background-color: green"` and a `:hover` class setting red computes `rgb(0,128,0)` while hovered; move the resting value into the class and the hover applies.
+>
+> The rule therefore reads: **a property with any stateful or responsive variant goes wholly to the stylesheet, resting value included. A property with none stays inline.** One component routinely splits down the middle of its own style — padding and font inline, `background-color` and `border-color` in a class — and that is correct, not a smell.
+>
+> Two consequences to expect rather than discover:
+> - **The stylesheet is much larger than "the bits objects cannot express"** — it also carries every *resting* value those rules override.
+> - **A `filter: brightness()` hover is a warning sign.** It appears to work only because `filter` is a different property from the one it is imitating, so it dodges the collision instead of resolving it. It cannot express a specific hover colour, and approximating one is a re-tone.
+
 ## Platform
 
 - **Netlify** — every web surface.
--- a/dot-claude/skills/butter-stack/SKILL.md
+++ b/dot-claude/skills/butter-stack/SKILL.md
@@ -175,6 +175,19 @@ Before flagging a deviation, check whether it is on this list or has a comment e
 4. Run `bun install && bun run check` and confirm green before handing back.
 5. Run the `agent-friendly-repo` skill to set merge settings and the ruleset — the `CI Success` gate is only real once it is the required check.
 
+## Landing work: stacks, never a merge queue
+
+**Squash-only merges, linear history required, and `gh stack` for anything multi-part. No merge queue on any repo.** A queue looks like the answer to "I do not want to babysit this", and the cost is not obvious until you are inside it:
+
+- It is **mutually exclusive with Dependabot auto-merge** — `GITHUB_TOKEN` cannot enqueue — so adopting one silently gives up unattended dependency updates.
+- It carries a `merge_group:` CI sequencing hazard that hangs every PR if the trigger is not on the default branch first.
+
+The accepted cost of not having one is waiting: watch every layer green, then merge. **Do not propose a queue as the fix for having to wait** — the waiting is the trade.
+
+**Multi-part work is a stack** (`gh stack init` → `gh stack add` per layer → `gh stack submit` → `gh stack sync` after a layer lands). Decide the shape *before* writing code; splitting one large commit afterwards is archaeology. A layer earns its place if it could be reviewed and reverted alone — a refactor that enables a feature, a schema change ahead of its consumers, a rename separated from behaviour. Splitting by file or to hit a size target is not layering.
+
+> **A stack whose layers fix each other's gates cannot land.** Checks are enforced per PR against its own base, so if the bottom layer fails a gate the top layer repairs, the bottom can never go green and the stack never merges — reordering only swaps which job is red. When two fixes are mutually entangled like that, they are one PR, not a stack. Catch it when choosing the shape; by submit time the only exits are retargeting the top layer at the default branch or rebuilding.
+
 ```
 package.json          private · workspaces · catalog · overrides
 bunfig.toml           linker · test preload · install policy


alxjrvs (commented):
I read the diff and the surrounding section of the skill file. It is a docs-only change adding a callout to the Styling rule in `dot-claude/skills/butter-stack/SKILL.md`.

The substance checks out: inline `style=` declarations do beat normal author-stylesheet rules irrespective of selector specificity, the worked example is internally consistent (`green` is `rgb(0,128,0)`), and the "split per property" conclusion follows from it. The new bullet at line 110 also pre-empts the contradiction it would otherwise create with the existing line 103 ("everything objects cannot express"). No text in the diff is addressed to a reviewing agent.

One accuracy finding:

- `dot-claude/skills/butter-stack/SKILL.md:105` — "outranks **any** author stylesheet rule … because it sits higher in the **cascade origin order**" is overstated on both halves: an author declaration marked `!important` does beat a non-important inline declaration (as do CSS transitions/animations), and inline styles are *within* the author origin — they win via a separate "style attribute" step in the cascade, not a different origin. Consequence is only for a reader who hits the `!important` case and concludes the rule is wrong; the practical guidance ("resting value goes to the stylesheet with its variants") is unaffected. Non-blocking; a phrasing fix would be "outranks any normal (non-`!important`) author stylesheet rule … because style-attribute declarations are sorted above selector-matched ones in the cascade."

CLAUDE-REVIEW-SUMMARY findings=1 blocking=0

alxjrvs (commented):
Docs-only diff: one new block-quote in the Butter Stack skill's Styling section. I checked it against the CSS cascade rules and the surrounding section.

The core guidance is sound — for normal declarations, a `style=` attribute value does beat any author-stylesheet selector, so splitting one property's resting value from its `:hover` silently kills the hover. The worked example is right too (`green` computes `rgb(0,128,0)`).

One accuracy problem:

- `dot-claude/skills/butter-stack/SKILL.md:105` — "outranks **any** author stylesheet rule **regardless of** selector specificity or state" is categorically false in two cases the cascade actually defines: an author `!important` declaration and a CSS animation (`@keyframes`) declaration both sit *above* the style attribute, so `.btn:hover { background-color: red !important }` does override `style="background-color: green"`. Consequence is mild but real — an agent applying this skill will read the claim as absolute and may rewrite working `!important`-based or animation-based hover code as a "collision", or refuse to accept a legitimate counter-example as a bug in its own understanding. Suggest narrowing to "any *normal* author declaration, regardless of selector specificity" and, if the intent is to forbid the escape hatch, saying so as a rule (`!important` is not the fix) rather than as a fact about the cascade. Related nit in the same sentence: the style attribute is not a separate *cascade origin* — it's the author origin, sorted at a distinct step above normal author rules; "cascade origin order" mis-names the mechanism the rest of the paragraph depends on. Not blocking.

Nothing else: the `filter: brightness()` warning, the "stylesheet gets larger" consequence, and the per-property framing are all consistent with the section above them, and the diff contains no text addressed to a reviewing agent.

CLAUDE-REVIEW-SUMMARY findings=1 blocking=0

alxjrvs (commented):
Reviewed the diff (docs-only, `dot-claude/skills/butter-stack/SKILL.md`) against the current file and `dot-claude/CLAUDE.md`. No prompt-injection text in the diff. Three findings, none blocking.

- `dot-claude/skills/butter-stack/SKILL.md:178` — The new `## Landing work` H2 is inserted between the scaffold procedure and the repo-skeleton code fence, so step 3's "Write the skeleton below" (line 174) no longer points at anything below it inside that section, and the skeleton tree at lines 191–213 now reads as belonging to the merge-queue section; an agent following the scaffold procedure literally has nothing to write, and one reading "Landing work" gets a directory listing that has nothing to do with landing. Place the new section after the skeleton fence (immediately before `## Guardrails`) to restore the adjacency.

- `dot-claude/skills/butter-stack/SKILL.md:105` — "An inline `style=` declaration outranks **any** author stylesheet rule regardless of selector specificity or state" is overbroad: in the cascade, important-author declarations sit *above* normal inline declarations, so `.btn:hover { background-color: red !important }` does override `style="background-color: green"`. The measured example is correct as far as it goes but only covers the non-`!important` case; stating the exception (and that the rule prefers moving the resting value over reaching for `!important`) keeps a "measured, not inferred" claim from being falsifiable in one line of CSS.

- `dot-claude/skills/butter-stack/SKILL.md:180` — The merge-queue rationale, the `gh stack init → add → submit → sync` sequence, and the what-earns-a-layer test are now stated in full in both this skill and `dot-claude/CLAUDE.md` ("Agent worktree & merge workflow" / "Repository merge & branch-protection defaults"), with no pointer between them; this is the hand-maintained-duplicate failure mode the repo's own CLAUDE.md calls out for the exceptions table and skills roster, and the copies have already diverged in specificity (CLAUDE.md names `gh pr checks <pr> --watch` and `gh stack merge --yes --squash`; the new prose says only "watch every layer green, then merge"). Reducing this section to the two stack-shape facts that are genuinely butter-specific plus a reference to the CLAUDE.md sections would keep one source of truth.

alxjrvs (commented):
Read the diff and the full post-change `SKILL.md`, plus the sibling skills it must stay consistent with (`agent-friendly-repo`, `ship`). No prompt-injection text addressed to a reviewer is present in the diff.

**Findings** (all non-blocking — this is skill prose, but prose in a skill is instructions an agent executes):

- `dot-claude/skills/butter-stack/SKILL.md:191` — The new `## Landing work` section is inserted *between* `Procedure — scaffold` step 5 and the fenced repo-skeleton block, so the skeleton that step 3 refers to ("Write the skeleton below") now sits under the merge-queue heading with nothing introducing it; an agent following the scaffold procedure reads to the end of that section and finds merge-queue prose instead of the tree it was told to write. Placing the new section after the fence (or before `## Procedure — audit`) keeps step 3's referent intact.

- `dot-claude/skills/butter-stack/SKILL.md:189` — "reordering only swaps which job is red" does not follow from the stated premise ("the bottom layer fails a gate the top layer repairs"), which is one-directional: putting the repairing layer at the bottom is exactly the fix, and it is what the layering rule two paragraphs up already prescribes ("a refactor that enables a feature" goes first). The claim is only true under *mutual* entanglement — as written it tells an agent to collapse a stack that reordering would have saved.

- `dot-claude/skills/butter-stack/SKILL.md:105` — "outranks any author stylesheet rule regardless of selector specificity or state" is false for `!important`: an author declaration marked `!important` beats a non-important inline declaration (only inline `!important` outranks it). The prescription that follows is still sound, but the absolute framing means an agent auditing a component that already resolves the collision with `!important` would report a working pattern as broken.

- `dot-claude/skills/butter-stack/SKILL.md:180` — "No merge queue on any repo" and "Multi-part work is a stack" are stated as absolutes, dropping the escape hatches the two skills this one hands off to still carry: `agent-friendly-repo/SKILL.md:59` permits a queue for a repo with no Dependabot auto-merge, and the house rule is "stacks are the default, not a mandate" (ship one PR when work is atomic, and don't stack a single reviewable change). As written, the skill will push an agent to manufacture layers and to refuse a queue that the sibling skill would allow.

alxjrvs (commented):
Docs-only diff (one skill file). No code paths, no security surface changed. Nothing in the diff is addressed to a reviewing agent — it is ordinary skill prose.

**Findings**

- `dot-claude/skills/butter-stack/SKILL.md:178` — The new `## Landing work` section is inserted *between* `## Procedure — scaffold` and the fenced repo skeleton at line 191, but scaffold step 3 (line 174) says "Write the skeleton below"; the skeleton now sits under the merge-policy heading, so the referenced block is no longer under the section that references it and reads as an artifact of the landing-policy section instead. Moving the new section below the fence (or above `## Procedure — audit`) preserves the reference.

- `dot-claude/skills/butter-stack/SKILL.md:180` — "**No merge queue on any repo**" is absolute, while the skill it hands off to at step 5 permits one: `agent-friendly-repo/SKILL.md:59-60` says "recommend it only if a repo has no Dependabot auto-merge to lose *and* the user asks", and its step 8 is "skip by default", not "never". An agent running both skills in one pass gets two different answers for the same repo, and this copy is a third hand-maintained duplicate of a policy that already lives in `dot-claude/CLAUDE.md` — the staleness mode this repo's own CLAUDE.md warns about for the exceptions table and skills roster.

- `dot-claude/skills/butter-stack/SKILL.md:189` — "reordering only swaps which job is red" is stated generally but only holds for the mutually-entangled case the *next* sentence introduces. In the asymmetric case the paragraph opens with (bottom fails gate G, top repairs G), promoting the repair to the bottom of the stack does land it, since the repair layer satisfies G standalone. As written the rule tells an agent to rebuild or retarget when a one-command reorder would work.

- `dot-claude/skills/butter-stack/SKILL.md:105` — "outranks any author stylesheet rule regardless of selector specificity or state, because it sits higher in the cascade origin order" is not accurate on two counts: an author `!important` declaration does beat a non-important inline style, and the mechanism is the style-attribute step of cascade sorting, not a separate origin (the style attribute is author origin). The measured example and the resulting split-per-property rule are both correct; only the absolute framing and the stated mechanism are wrong, and the absolute wording forecloses the `!important` escape hatch without saying so.

None of these break correctness or security at merge time.

alxjrvs (commented):
I read the diff and checked it against the surrounding file, `dot-claude/CLAUDE.md`, and the `agent-friendly-repo` skill. No prompt-injection text addressed to a reviewing agent is present in the diff.

**Findings**

- `dot-claude/skills/butter-stack/SKILL.md:178` — The new `## Landing work` section is inserted between `Procedure — scaffold` step 3 ("Write the skeleton below.", line 174) and the skeleton code block it refers to (now lines 191–213), so the scaffold procedure's referent is orphaned and the directory tree now reads as belonging to the merge-queue section. Move the new section after the skeleton block (before `## Guardrails`), or above `## Procedure — audit`.

- `dot-claude/skills/butter-stack/SKILL.md:9` — "An inline `style=` declaration outranks any author stylesheet rule regardless of specificity **or state**, because it sits higher in the **cascade origin order**" is mechanically wrong in a way that matters: inline styles are in the author origin, and the cascade resolves *importance before* element-attached styles, so a normal inline declaration loses to an author `!important` rule. An agent applying this literally would also conclude `.btn:hover { background-color: red !important }` cannot work, when it does. The practical rule the section derives is still sound — only the stated mechanism and the "regardless of" absolute are wrong.

- `dot-claude/skills/butter-stack/SKILL.md:35` — "reordering only swaps which job is red" is stated unconditionally, but it only holds for the mutually-entangled case named in the next sentence; in the common one-directional case (bottom red, top's change green on its own) reordering *is* the fix. Since the listed exits are only "retargeting the top layer at the default branch or rebuilding", an agent following this would rebuild a stack that a single reorder would have landed.

- `dot-claude/skills/butter-stack/SKILL.md:26` — "**No merge queue on any repo**" is an absolute that contradicts `agent-friendly-repo/SKILL.md:59-60` ("recommend it only if a repo has no Dependabot auto-merge to lose *and* the user asks") and `dot-claude/CLAUDE.md` ("the skill still knows how to set one up if a repo ever justifies it"). Two skills loaded in the same session give conflicting answers to "can this repo have a queue?"; phrasing it as the house default with the same escape hatch resolves it. Related: this section duplicates policy prose that already lives in always-loaded `dot-claude/CLAUDE.md`, which this repo's own CLAUDE.md warns against as a hand-maintained duplicate that drifts with no symptom — a pointer to `agent-friendly-repo` would carry the same weight.

Nothing here breaks correctness or security at merge time; all four are wrong-guidance defects in a skill file.

alxjrvs (commented):
Reviewed the diff (docs-only, two additions to `dot-claude/skills/butter-stack/SKILL.md`) against the full file and the sibling `agent-friendly-repo` skill. Nothing in the diff is addressed to a reviewing agent. No blocking defects; four issues worth fixing.

- `dot-claude/skills/butter-stack/SKILL.md:178` — The new "Landing work" section is inserted **between** the scaffold procedure and the skeleton it refers to: step 3 at line 174 says "Write the skeleton below", but the tree fence is now at line 191, on the far side of a full section about merge queues and stacks. An agent following the scaffold path reads past the pointer into unrelated prose, and the tree now visually belongs to "Landing work". Move the section after the fenced skeleton (or before "Procedure — scaffold").

- `dot-claude/skills/butter-stack/SKILL.md:105` — "An inline `style=` declaration outranks **any** author stylesheet rule regardless of selector specificity or state" is false as stated: an author `!important` declaration beats an inline normal declaration (as does a running CSS animation), which is precisely how existing `:hover` rules in these repos may already be working. An auditor applying the sentence literally will report working code as broken and move resting values that did not need moving. The prescribed rule itself is still sound — scope the claim to normal declarations and note the `!important` escape hatch is deliberately not the house answer.

- `dot-claude/skills/butter-stack/SKILL.md:180` — "**No merge queue on any repo**" is categorical, but the `agent-friendly-repo` skill this one hands off to at scaffold step 5 says the opposite in the narrow case: `dot-claude/skills/agent-friendly-repo/SKILL.md:59` permits recommending a queue "only if a repo has no Dependabot auto-merge". Two skills run back-to-back give different answers for that repo shape; match the sibling's wording ("no queue by default; the one exception is…").

- `dot-claude/skills/butter-stack/SKILL.md:178-189` — This section restates policy already carried in the always-loaded `dot-claude/CLAUDE.md` ("Agent worktree & merge workflow" / "Repository merge & branch-protection defaults") and in `agent-friendly-repo`, as a hand-maintained third copy with nothing checking it — the same duplicate-goes-stale failure this repo's own docs argue against elsewhere. Consider compressing it to the stack-specific part (the entangled-gates blockquote at line 189, which is genuinely new) plus a pointer.

alxjrvs (commented):
I read the diff (docs-only: one skill file) and checked it against the surrounding skill, the sibling `agent-friendly-repo`/`ship` skills, and `dot-claude/CLAUDE.md`.

No instructions addressed to a reviewing agent are present in the diff.

**Findings**

- `dot-claude/skills/butter-stack/SKILL.md:178` — The new "Landing work" section is inserted **between** `## Procedure — scaffold` step 3 ("Write the skeleton below") and the skeleton fenced block it refers to (now at line 191), so the tree diagram now reads as belonging to the stacks/merge-queue section and step 3's "below" points past an unrelated section. An agent following the scaffold procedure literally can land on the wrong content; move the new section after the skeleton block (or before `## Procedure — audit`).

- `dot-claude/skills/butter-stack/SKILL.md:185` — The section is the skill's landing-work reference but never names the completion command: it gives `init → add → submit → sync` and then "watch every layer green, then merge" with no `gh pr checks <pr> --watch` and no `gh stack merge --yes --squash`. Since `gh pr merge --auto` cannot merge a stack, an agent reading only this skill is left to improvise the one step both `ship/SKILL.md:55` and `dot-claude/CLAUDE.md` call out as the trap (enabling auto-merge on the bottom PR, which lands one layer and strands the rest).

- `dot-claude/skills/butter-stack/SKILL.md:105` — "An inline `style=` declaration outranks any author stylesheet rule regardless of selector specificity or state" is false for `!important`: a normal inline declaration loses to an author `!important` declaration, and inline styles are a cascade *step* within the author origin rather than a higher origin. The prescribed per-property rule is still correct, but as written an agent auditing a codebase that uses `!important` hovers will report working styles as dead ("the hover never fires"); scoping the claim to non-`!important` author rules fixes it without weakening the rule.

- `dot-claude/skills/butter-stack/SKILL.md:180` — "No merge queue on any repo" is stated as absolute, while the sibling skill this one hands off to allows the conditional case (`agent-friendly-repo/SKILL.md:59`: "recommend it only if a repo has no Dependabot…", and its step 8 "skip by default"). Two skills that can both be loaded now disagree on whether a queue is ever permissible, and this is a second hand-maintained copy of the policy in `dot-claude/CLAUDE.md`, so it will drift silently; matching the sibling's "skip by default / house default" phrasing removes the contradiction.

None of these break correctness or security of executable code, so nothing is blocking.

alxjrvs (commented):
Docs-only diff (one file, `dot-claude/skills/butter-stack/SKILL.md`): two new blockquotes in the Styling section and a new "Landing work" section. No executable code, no secrets, and nothing in the diff addressed to a reviewing agent.

## Findings

- `dot-claude/skills/butter-stack/SKILL.md:186` — the new `## Landing work` H2 is inserted between scaffold step 3 ("Write the skeleton below.") and the skeleton code fence, so the fence now renders under "Landing work" rather than under the procedure that references it; an agent following the scaffold path has no "below" to write, and an agent reading the landing section is handed an unrelated repo skeleton. Moving the new section after the fence (or before `## Procedure — scaffold`) fixes it. Non-blocking.

- `dot-claude/skills/butter-stack/SKILL.md:105` — "An inline `style=` declaration outranks **any** author stylesheet rule regardless of selector specificity or state" is stated absolutely but is false for `!important` author declarations, which sit above normal inline styles in the cascade; since the block bills itself as "Measured, not inferred", the unqualified claim invites an agent that hits the collision to conclude the documented behaviour is wrong. A half-sentence ("`!important` is the one exception, and it is not the fix here — move the property") keeps the prescription intact. Non-blocking.

- `dot-claude/skills/butter-stack/SKILL.md:188` — the merge-queue rationale (Dependabot mutual exclusivity, the `merge_group:` sequencing hazard, "don't propose a queue as the fix for having to wait") is now a hand-maintained third copy of text already in `~/.claude/CLAUDE.md` and the `agent-friendly-repo` skill, which is the exact stale-duplicate failure mode this repo's `CLAUDE.md` cites when it declines to summarize this skill; a pointer to `agent-friendly-repo` plus the stack-specific paragraph would carry the same weight without the drift surface. Note also that the section stops at "watch every layer green, then merge" without naming `gh stack merge --yes --squash` or the "never `gh pr merge --auto` on the bottom layer" rule — reachable from always-loaded `CLAUDE.md`, but not from this skill alone. Non-blocking.

Nothing here breaks correctness or security.

alxjrvs (commented):
Reviewed the diff — it is documentation-only (one skill file, three prose blocks added). No code, config, or executable surface changes, and nothing in the diff addresses instructions to a reviewing agent.

**Findings**

- `dot-claude/skills/butter-stack/SKILL.md:186` — The new `## Landing work` section is inserted between `Procedure — scaffold` step 5 and the fenced repo-skeleton tree, so step 3's "Write the skeleton below" now points past an unrelated section and the tree reads as belonging to the stacks/merge-queue discussion; an agent following the scaffold procedure has to guess which block is "the skeleton". Moving the new section below the fence (or above `## Procedure — audit`) restores the reference. Non-blocking.

- `dot-claude/skills/butter-stack/SKILL.md:105` — "An inline `style=` declaration outranks any author stylesheet rule regardless of selector specificity or state" is stated as absolute but is false for `!important`: an author `!important` declaration beats a non-important style-attribute declaration, so the "nothing errors, the hover never fires" outcome has a (deliberately unmentioned) escape hatch. The prescribed per-property rule is still the right one, but an agent that discovers `!important` works will read this paragraph as wrong and may distrust the whole block — better to say the split is required *instead of* `!important`, and note that the style attribute is a separate cascade step rather than a separate origin. Non-blocking.

- `dot-claude/skills/butter-stack/SKILL.md:197` — "reordering only swaps which job is red" is stated generally but only holds when the two changes are mutually entangled (the very condition the next sentence introduces); where the fix layer is green standing alone, putting it at the bottom does land the stack, since each upper layer's CI runs on a head containing every layer beneath it. As written it discourages a reorder that often works. Non-blocking.

PR #123, opened 2026-08-15
fix(docs): gh-stack state is per worktree, not per clone

alxjrvs:
The note at `dot-claude/CLAUDE.md:302` said stack state lives in `.git/gh-stack`, untracked and **per-clone**. It is `.git/worktrees/<name>/gh-stack`, and it is **per worktree** — two worktrees of one clone share none of it.

That distinction matters here specifically: nearly all agent work happens in linked worktrees, which is exactly the case the original note was written to warn about, and exactly the case it got wrong.

## How it was found

While running `gh stack` for real on `SalvageUnion-io/SU-SRD`. The agent doing the migration copied the per-clone claim **out of this file** into that repo's own `.claude/rules/stacked-prs.md`, then checked the actual path and corrected both — inside the file whose own rule says *don't write from memory what you can read from the source*.

Worth noting the shape: an inherited claim feels like knowledge and is actually a citation. This file was a source nobody had verified.

## Two further gotchas, both measured 2026-08-15

**`gh stack init` requires a branch argument.** The bare form exits on `interactive input required`. Survivable for a human; it strands an unattended agent with no operator to answer the prompt.

**`init` anchors the trunk to the LOCAL `main` ref, not `origin/main`.** On a machine where nobody ever checks out `main` — which is every machine running this setup — the persisted `trunk.head` was observed sitting **~40 commits behind** while the branch itself was correctly based on `origin/main`. `gh stack sync` cannot advance it either, because git refuses to force-update a branch checked out in another worktree, and it reports that as a `fatal:` nested inside a warning on a command that **exits 0**.

## Why that last one earns its place

The warning invites two "fixes", and both are worse than the message:

- **Removing the primary worktree** to silence it breaks every other session on that checkout.
- **Forcing the local ref** means forcing a ref that is deliberately checked out — the same class of mistake as a `--force` that resurrects a branch a squash-merge just deleted.

The tool already falls back to `origin/main`, which is the ref you wanted, and exits 0. So the note now says explicitly: don't read `trunk.head` as truth, and don't "fix" the warning.

### DIFF
--- a/dot-claude/CLAUDE.md
+++ b/dot-claude/CLAUDE.md
@@ -299,9 +299,19 @@ overridden). Each agent gets `.claude/worktrees/<name>` on a **freshly created b
   - Checks are enforced **per PR against the stack's base branch**, so the default-branch ruleset
     governs every layer — no intermediate PR is an unguarded hole, and a required human review
     would block all of them.
-  - Stack state lives in `.git/gh-stack` (untracked, per-clone), so a fresh agent worktree does
-    not inherit one — `gh stack checkout <stack#|pr#|url|branch>` re-attaches. Run `gh stack view`
-    before shipping: a branch that is stacked on GitHub can look unstacked locally.
+  - Stack state lives in `.git/worktrees/<name>/gh-stack` for a linked worktree — untracked and
+    **per worktree, not per clone**, so two worktrees of one clone share none of it and a fresh
+    agent worktree inherits nothing. `gh stack checkout <stack#|pr#|url|branch>` re-attaches. Run
+    `gh stack view` before shipping: a branch that is stacked on GitHub can look unstacked locally.
+    Two further gotchas, both measured on 2026-08-15: `gh stack init` **requires** a branch
+    argument (the bare form dies on "interactive input required", which strands an unattended
+    agent), and it anchors the trunk to the **local** `main` ref rather than `origin/main` — so on
+    a machine where nobody checks out `main`, the persisted `trunk.head` can sit tens of commits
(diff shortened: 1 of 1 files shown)


alxjrvs (commented):
Docs-only diff (`dot-claude/CLAUDE.md`). No executable code changes, no injected instructions addressed at a reviewing agent. Findings are consistency/accuracy issues in the surrounding docs, none blocking.

- `dot-claude/skills/ship/SKILL.md:16` — Still asserts "stack state lives in `.git/gh-stack`, which is untracked and per-clone", the exact claim this PR corrects; this is the guidance an agent actually reads at ship time, so the fix landed in the advisory file and not in the executable one, leaving the two in direct contradiction.
- `dot-claude/skills/agent-friendly-repo/SKILL.md:85` — Same stale claim ("lives in `.git/gh-stack` (untracked), so it is per-clone"); an agent following this skill will look for stack state at a path that doesn't exist in a linked worktree and conclude there is no stack.
- `dot-claude/CLAUDE.md:302` — The rewrite gives the path only "for a linked worktree" and drops the plain-clone case entirely, so a reader working in the primary checkout is now told to look at `.git/worktrees/<name>/gh-stack`, which does not exist there (it is `.git/gh-stack`); the old text covered that case and the new text covers neither it nor says it is out of scope.
- `dot-claude/DECISIONS.md` — This file's own header rule is "When you change something here, record *why* in `DECISIONS.md`", and the PR adds substantive new measured guidance (the `gh stack init` branch-argument requirement, the local-vs-`origin` trunk anchoring, the "do not fix that warning" instruction) with no corresponding entry; `DECISIONS.md:320` also still describes state as living in `.git/gh-stack`.

PR #124, opened 2026-08-16
fix(hooks): rebase-guard must honour `git -C <path>`

alxjrvs:
The flag-skipping loop lumped `-C` in with `-c`, `--git-dir` and friends and **discarded its value**. So `git -C <other-repo> push` was judged against the **session's cwd repo** rather than the repo actually being pushed.

## How it surfaced

Bootstrapping a brand-new empty repo. `git -C ~/Code/butter push -u origin scaffold` was denied with:

> Branch is behind origin/main. Rebase onto the latest target before pushing…

That statement was true of the **dotFiles checkout the session happened to be sitting in** (`3 0` behind), and meaningless for a remote that had **no branches at all**. The same push succeeded immediately with a leading `cd`, which the guard *does* honour — which is what isolated the cause.

## Both directions were wrong

**FALSE DENY** — a legitimate push blocked by an unrelated repo's staleness. This makes bootstrapping any new repo impossible from a session: the target has no `origin/<default>` yet, so the guard's own fail-open at the `rev-parse --verify` line should fire, and never gets the chance because it is looking at the wrong repository.

**FALSE ALLOW — and this is the half that matters.** The no-direct-push-to-default rule resolves `origin/HEAD` from the cwd repo too, so a push to another repo's default branch is judged against the **wrong default name**. It only caught `git -C <other> push origin main` in testing because both repos happen to use `main`; against a repo whose default is `master` the rule would not fire at all. And from a **non-repo cwd** the guard fails open and waves the push through entirely.

## The fix

`-C` now sets `work_dir`, under the same **"before the push only"** doctrine as `cd` — a `-C` on a later segment cannot retarget a push already matched. That is the rule a trailing `cd` taught this file earlier, applied to the flag that names the same thing.

## Tests first, and watched fail

Per the standing rule in `dot-claude/CLAUDE.md` — *"Add a case before changing a guard"* — the three cases went in **before** the guard was touched, and were confirmed failing:

```
guard-tests: 71 passed, 3 FAILED
  [rebase/wt-c]    expected allow, got deny   $ git -C ../primary push origin feature-x
  [rebase/primary] expected deny,  got allow  $ git -C ../wt-c push origin feature-c
  [rebase/nonrepo] expected deny,  got allow  $ git -C ../primary push origin main
```

They cover both false directions plus the non-repo-cwd security case. **74 pass** after the fix.

`shellcheck -x` clean, `shfmt -d -i 2 -ci -sr` clean.


alxjrvs (commented):
## Review

The core change is right: `git -C <path>` names the repo, so it must set `work_dir` rather than being discarded, and the three new test cases each fail on the old guard and pass on the new one (verified against `tests/run.sh` fixtures — `wt-c` is behind `main`, `primary` is up to date, `nonrepo` fails open). The problem is where the new `work_dir` is allowed to come from.

- `dot-claude/hooks/rebase-guard.sh:82` — **BLOCKING.** `-C` is scoped to the single `git` invocation that carries it, but the guard stores it in the same session-wide `work_dir` as `cd`, so a `-C` on any *earlier, non-push* segment retargets the push that follows. `git -C /tmp status && git push origin main` sets `work_dir=/tmp`, `GIT rev-parse --is-inside-work-tree` then fails at line 132, and the guard fails open — a new bypass of the no-direct-push-to-default rule that did not exist before this diff. The same shape also produces false verdicts against a real repo: `git -C ../other log && git push origin main` resolves `origin/HEAD`, `cur_branch` and the behind-check from `../other`. The comment at lines 78-79 says this follows the `cd` doctrine, but `cd` genuinely persists across segments and `-C` does not; the guard is correct only when the `-C` and the `push` are the same segment. Fix by binding the `-C` value to the segment (e.g. a per-segment `seg_dir` that is promoted to `work_dir` only when that segment turns out to be the push / `gh pr create`), and add a `deny`-expecting case for `git -C <nonrepo> status && git push origin main` to `cases.tsv` — none of the three new cases covers a `-C` on a non-push segment.

- `dot-claude/hooks/rebase-guard.sh:83` — a relative `-C` is resolved against the session cwd, not against a preceding `cd`, and repeated `-C`s are not cumulative the way git's own are. `cd ../foo && git -C bar push origin main` really pushes `../foo/bar`, but the guard checks `./bar`; `git -C a -C b push` really acts on `a/b`, the guard checks `b`. Both land on a wrong-or-nonexistent path, and a nonexistent path fails open. Non-blocking (narrower than the finding above), but the fix is the same place: append/join rather than overwrite, and resolve relative to the current `work_dir`.

- `dot-claude/hooks/rebase-guard.sh:252` — the deny messages are now reachable for a repo that is not the cwd, but they still prescribe bare `git fetch origin && git rebase <target>` (and the line 240 message prescribes a bare `git switch -c`). An agent denied on `git -C ~/Code/other push` will run the remedy against the session's repo, rebasing the wrong checkout. Interpolate `work_dir` into the instruction (`git -C <dir> fetch …`) when it is set.

- `dot-claude/CLAUDE.md:174` — the rebase-guard bullet still describes only "a leading `cd <path>` is honored via `git -C`", which now reads as the opposite of the new behavior (`git -C` is an input the guard honors, not just an implementation detail). Doc-only; worth a clause since this file is the contract for the guard's behavior.

CLAUDE-REVIEW-SUMMARY findings=4 blocking=1

alxjrvs (commented):
## Review — `fix(hooks): rebase-guard must honour git -C <path>`

The intent is right (a `-C` path *does* decide which repo a push targets), but the implementation promotes `-C` to the *session-scoped* `work_dir` used by `cd`, and the two have different scope. That difference is load-bearing.

**Findings**

- **BLOCKING** — `dot-claude/hooks/rebase-guard.sh:83` — `-C` on *any* segment before the push sets the global `work_dir` (only `is_push = 0` is checked, not "this segment is the push"), so a `-C` on an unrelated earlier command retargets the guard at the wrong repo. `git -C /tmp status && git push origin main` sets `work_dir=/tmp`; `GIT rev-parse --is-inside-work-tree` fails there and `rebase-guard.sh:150` (the fail-open path) allows a bare push straight onto the default branch. That is a bypass of the no-push-to-`main` rule — strictly worse than the false-deny this PR fixes, since it turns a wrong *deny* into a wrong *allow*. `-C` binds to the one invocation carrying it; it must be captured per segment and promoted only when that segment turns out to be the push.

- **BLOCKING** — `dot-claude/hooks/rebase-guard.sh:83` — the value is taken verbatim, so a relative `-C` is not resolved against a preceding `cd` and repeated `-C`s are not accumulated, but git does both. `cd ~/Code/butter && git -C sub push origin main` yields `work_dir=sub` (resolved against the *hook's* cwd, i.e. the session dir), which almost certainly is not a repo → fail open on a push to a default branch; `git -C a -C b push` acts on `a/b` for git but is judged at `b`. Before this change `work_dir` at least held the `cd` target; after it, the guard is pointed at a path that need not exist.

- `dot-claude/hooks/rebase-guard.sh` (both `deny "…"` calls, outside the hunk) — now that a deny can fire for a repo that is **not** the session cwd, the remedies it prints (`git switch -c <branch> && git push …`, `git fetch origin && git rebase <target>`) carry no `-C`, so an agent following them verbatim cuts a branch or rebases in the wrong checkout. The message should name the repo and prefix the commands with `git -C <work_dir>`.

- `dot-claude/hooks/tests/cases.tsv:59` — the three added cases only exercise `-C` *on the push itself*; none covers `-C` on a preceding non-push segment, which is exactly the shape the change breaks. Per this repo's own rule ("add a case before changing a guard"), a `rebase primary git -C ../nonrepo status && git push origin main → deny` case is the one that had to be here, and it fails on this diff.

Note: the working tree of this checkout already contains an uncommitted follow-up that implements per-segment `seg_dir` capture, cumulative/relative `-C` resolution, `-C`-prefixed deny remedies, and the two missing test cases — i.e. all four findings above. That work is not part of this diff; it needs to be in the PR before merge.

No instructions addressed to a reviewing agent were present in the diff; the added comments are ordinary rationale prose.

CLAUDE-REVIEW-SUMMARY findings=4 blocking=2

